PR #77, opened 2016-08-02
Add country bbox's

olafveerman:
Generated using: https://gist.github.com/olafveerman/1496a45735f9a79ff2d76584e1cc7621
Any use in committing the above script to the repo. In `/tools` for example?


danielfdsilva:
I don't think so. It's not like we'll be generating this very often

PR #91, opened 2016-08-22
Fixes

ascalamogna:


danielfdsilva:
@ascalamogna Could you please review the images resolution? Not sure what's the idea retina resolution but right now they're too big. And then run them by a compressor. 6+ MB is definitely not something we want. :)
Everything else looks great!

PR #106, opened 2016-09-01
Deploy V2

danielfdsilva:


jflasher:
Oh, way to use multiple assignees. :) I'll merge if it passes and we can take a look at it on staging.

PR #107, opened 2016-09-01
Deploying v2

jflasher:
I feel really bad I didn't squash the commits.


olafveerman:
@jflasher It should be good to go! `keep-theirs` ftw

PR #123, opened 2016-09-22
Encode location names

### LINKED ISSUE #122: unique urls have spaces
The unique url functionality for pages like the 'comparisons' one is awesome. The only thing is, because spaces appear, the url will not link properly in twitter or facebook. To get around that, I tried shortening it with tinyurl.com, but it will not accept it, either. 
bitly.com will, but it's an unideal workaround.

Here's an example url: 
https://openaq.org/#/compare/Camden Kerbside/MNB/US Diplomatic Post: Pristina?parameter=pm25&_k=huwix6

danielfdsilva:
Fix #122 

Quite interesting behavior this one. Firefox automatically encodes the url when you copy it, but chrome doesn't. I used `encodeURIComponent` to solve this.

On firefox the url will still be shown with spaces but when you copy it will be encoded. i.e. the spaces will be repalced with `%20`. On chrome the url will be encoded in the url bar from the start.

Could you see if it behaves as expected? @jflasher @RocketD0g

### DIFF
--- a/app/assets/scripts/components/location-card.js
+++ b/app/assets/scripts/components/location-card.js
@@ -40,7 +40,7 @@ var LocationCard = React.createClass({
         <div className='card__contents'>
           <header className='card__header'>
             <p className='card__subtitle'>Updated <strong>{updated}</strong></p>
-            <h1 className='card__title'><Link to={`/location/${this.props.name}`} title={`View ${this.props.name} page`}>{this.props.name}</Link> <small>in {this.props.city}, {this.props.countryData.name}</small></h1>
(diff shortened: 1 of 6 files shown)


jflasher:
Thanks for taking a look so quickly @danielfdsilva , will check this out now.

jflasher:
When starting at /compare, adding a new location doesn't update nav bar with encoded url.

1. http://localhost:3000/#/compare
1. Add location -> Ethiopia -> Addis -> US Dip post

nav bar url not encoded when location is added.

danielfdsilva:
@jflasher :boom: done!

RocketD0g:
Wow, that was so fast! Thanks! 💃

PR #128, opened 2016-10-25
Prevent Jeet from breaking on next build

olafveerman:
See: https://github.com/mojotech/jeet/issues/504

cc @danielfdsilva


jflasher:
@olafveerman @danielfdsilva looks like `eslint-plugin-promise` needs updated? Fun with packaging!

olafveerman:
Just want to note that my intentions with the original PR were good.

Issue #139, opened 2016-11-03
Map coloring and point size 


ascalamogna:
We got a few comments on the current map colors not being bright enough and large enough, and preferences for the old map color scheme where expressed, where it was clearer where hot spots/cold spots were.

I personally have a strong preference for the old color scheme, or one similar to it, and to make the points larger and more distinct. I think it is very important to see as soon as you look at the map on a global level which places a) have any data, and b) have high versus low levels of pollution. One external factor with the coloring: while we are not using any particular AQI scale purposely, nearly all scales for pollutants - for better or worse - well-constructed or otherwise - use a red to blue/purple scale. People who look at air pollution maps are just used to seeing that and other significantly differing schemes can look jarring.

DylanMoriarty:
Ok. So, going back to a color scheme that's closer to the original...

*For these screenshots I bumped up the overall radius of each point a pixel for the low zoom levels.

1.
<img width="1157" alt="screenshot 2016-11-04 13 34 17" src="https://cloud.githubusercontent.com/assets/9341928/20016171/ce8bccaa-a294-11e6-93b5-483142637d68.png">

2. ( Old colors )
<img width="1115" alt="screenshot 2016-11-04 13 51 05" src="https://cloud.githubusercontent.com/assets/9341928/20016413/c1dd348e-a295-11e6-90ae-7837278f286e.png">

3. ( Tweaked older colors )
<img width="1119" alt="screenshot 2016-11-04 14 03 25" src="https://cloud.githubusercontent.com/assets/9341928/20016820/7eb7485a-a297-11e6-8cd1-cc439d07bed1.png">

4. ( No red )
<img width="1023" alt="screenshot 2016-11-04 14 12 53" src="https://cloud.githubusercontent.com/assets/9341928/20017171/d4a972be-a298-11e6-9036-8aeb4335340c.png">

For me, I dig 1# & 3#, and feel like the first does the better job of not making it look like there's a middle value or turning point for the AQ values. I know we'd discussed not using red for the areas with the most pollutants, so maybe a happy medium would be just dulling them a bit more.

@ascalamogna

ascalamogna:
@DylanMoriarty I like 1 and 3 as well. What do you think @olafveerman?

olafveerman:
I like 3 the most. What 'throws' me off on 1, is that the stronger green has a higher value than the blue. Instinctively, if I look at the map, I think that the green dots are the lower values.

@DylanMoriarty

[...]

PR #150, opened 2016-11-07
Update/point colors

DylanMoriarty:
Makes a new color scale for the point & legend features, satisfying #139. 

Any of @olafveerman @ascalamogna @jflasher for review.

### DIFF
--- a/app/assets/scripts/utils/colors.js
+++ b/app/assets/scripts/utils/colors.js
@@ -5,15 +5,15 @@ import { round } from '../utils/format';
 
 // Colors for the legend and point fills
 const mapColors = [
-  { label: '0', color: '#b5ebb1' },
-  { label: '10', color: '#97d8b5' },
-  { label: '20', color: '#6dc9c1' },
-  { label: '30', color: '#4db3d2' },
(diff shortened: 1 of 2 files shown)


ascalamogna:
@DylanMoriarty did you update the point size too in this PR? I don't see anything in the commits?

DylanMoriarty:
@ascalamogna nope. I can kick up a slight increase if you think it's still necessary with the color change.

ascalamogna:
@DylanMoriarty Let's give it a slight increase to see how people respond since it's being requested. Unless you think it looks crazy?

DylanMoriarty:
screenshot 2016-11-11 09 41 41

It gets a busier in dense area's, and that masks a few of the smaller points, but that was an issue anyway because the points aren't opaque. The color is indeed clearer, so I think it's better.

For comparison:
screenshot 2016-11-11 10 27 00

danielfdsilva:
@RocketD0g are you ok with the new color scheme?

olafveerman:
@danielfdsilva I think this is good to go. See: https://github.com/openaq/openaq.org/issues/139#issuecomment-258581660

PR #152, opened 2016-11-16
Sources information

danielfdsilva:


danielfdsilva:
There are still a couple of problems with some locations. For example:

This location https://api.openaq.org/v1/locations?location=SPARTAN%20-%20Tsinghua%20University has the source `Spartan`, but it doesn't exist in https://api.openaq.org/v1/sources

@jflasher is this something that can be added, or do we need to figure out a way to handle missing sources?

danielfdsilva:
There are still a couple of problems with some locations. For example:

This location https://api.openaq.org/v1/locations?location=SPARTAN%20-%20Tsinghua%20University has the source `Spartan`, but it doesn't exist in https://api.openaq.org/v1/sources

@jflasher is this something that can be added, or do we need to figure out a way to handle missing sources?

jflasher:
@danielfdsilva We can/should add these. I think the missing ones will be the original Indian ones and Spartan. I think I can add these in this evening, will that work?

danielfdsilva:
:+1:

jflasher:
Deploying the unused sources info nowish.

danielfdsilva:
@olafveerman can you please test and merge?

jflasher:
These are looking good, going to merge in. 💥

Issue #169, opened 2017-01-03
Facebook social sharing not working properly


jflasher:
When clicking Share -> Facebook, it goes to an error page (it's final url is https://www.facebook.com/location/Zinnwald?_k=tuiuxy). I have made sure I'm logged into FB before clicking share button.

Happens on Chrome 55.0.2883.95 on my MBP.

![](https://www.dropbox.com/s/i1tfk3zwfh5jrcy/Screenshot%202017-01-03%2010.24.32.png?dl=1)

aicarlson:
Confirming that this is an issue. For me though, the final URL is https://www.facebook.com/compare/

I'm looking into this.

Issue #168, opened 2017-01-03
Twitter sharing not including links properly


jflasher:
When clicking Share -> Twitter for a link like https://openaq.org/#/location/Zinnwald?_k=tuiuxy, the Twitter modal pops up, but the link it includes is only https://openaq.org. 

I would expect this to include rest of link. Final Twitter url for modal is https://twitter.com/intent/tweet?url=https%3A%2F%2Fopenaq.org%2F&original_referer=https%3A%2F%2Fopenaq.org%2F#/location/100%20ail?_k=d3m3ix, this may be an encoding issue when sending to Twitter?

Happens on Chrome 55.0.2883.95 on my MBP.

PR #170, opened 2017-01-05
Encodes the current URL for sharing

andrewicarlson:
This resolves https://github.com/openaq/openaq.org/issues/169 and https://github.com/openaq/openaq.org/issues/168 by encoding the current URL such that query strings don't throw off the social property consumption.

### DIFF
--- a/app/assets/scripts/components/share-btn.js
+++ b/app/assets/scripts/components/share-btn.js
@@ -10,7 +10,7 @@ var ShareBtn = React.createClass({
 
   render: function () {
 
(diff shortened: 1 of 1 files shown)


aicarlson:
Please note that if testing locally you won't be able to submit a URL with the root domain as localhost:{{PORT}} so I would recommend replacing `window.location.href` on line `13` with `https://openaq.org/#/compare/JOHN/40GK09%20-%20GENK?parameter=pm25&_k=27lnfl`

jflasher:
Thanks @aicarlson, looks good! 💥

PR #184, opened 2017-03-21
Show random community project

danielfdsilva:
I think the best way to implement this is to pick the project on page load, and use it throughout the session.


jflasher:
Thanks @danielfdsilva ! Will merge once tests pass.

PR #215, opened 2017-07-06
Fix error on locations with no coordinates.

### LINKED ISSUE #183: Error on location pages when no coordinates present
An example is https://openaq.org/#/location/Maninagar

We should gracefully handle this since a number of locations in the system do not have associated coordinates.

meiamsome:
Fixes #183.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ cache:
 before_install:
 - chmod +x ./.build_scripts/lint.sh
 script:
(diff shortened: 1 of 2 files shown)


meiamsome:
@jflasher Okay, I did 1 & 3, but 2 was because I mistakenly used `this.props.location.coordinates` instead of `loc.coordinates` in the if statement, so I fixed that instead.

PR #256, opened 2018-03-20
Allow build & serve commands to run on Windows (#255)

### LINKED ISSUE #255: Build & serve commands fail on Windows
When following the README steps to run the project locally on a Windows machine, after a successful `npm install`, `npm run serve` and `npm run build` commands both fail.

The relevant error message is
```
'DS_ENV' is not recognized as an internal or external command,
operable program or batch file.
```

Troubleshooting leads to [this SO question](https://stackoverflow.com/q/11928013/4167438); the Linux format of `DS_ENV=development gulp serve` will fail on Windows. [This answer](https://stackoverflow.com/a/37131308/4167438) seems like the most robust solution, to simply install a dev dependency, [`cross-env`](https://www.npmjs.com/package/cross-env), that will make the commands cross-platform. I'll give that a shot

jackkoppa:
Adds the [`cross-env`](https://www.npmjs.com/package/cross-env) dev dependency, and prefixes Linux-style npm commands with `cross-env`. This allowed the commands to work correctly locally on Windows; presumably if the CI build continues to pass, it remains successful on Linux as well.

fixes #255

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
   "homepage": "https://openaq.org",
   "scripts": {
     "postinstall": "[ -f app/assets/scripts/config/local.js ] || echo 'module.exports = {};' > app/assets/scripts/config/local.js",
-    "serve": "DS_ENV=development gulp serve",
-    "build": "NODE_ENV=production gulp",
+    "serve": "cross-env DS_ENV=development gulp serve",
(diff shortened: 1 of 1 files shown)


jflasher:
Thanks @jackkoppa!

PR #268, opened 2018-10-31
More detailed source attribution

CamLatimer:
This update enables extra attribution sources to be shown in the view for individual locations. It lists the source name, description, and contact info of the original as it did before.  It just adds the source names  and url in the attributions field of the api response in order to list other sources if there are others.  Tweaked app/assets/scripts/actions/measurements to get the attribution info in the api respsone and use it in app/assets/scripts/views/location. Also tweaked the jsx and app/assets/styles/location/_content to show the source names in one column and keep the description and contact info in another as well as add 'Sources:' as a section header.

Glad to look through the code and see if I could chip in.  Hope it helps a bit.  Let me know what you think, if anything needs to change, etc.

### DIFF
--- a/app/assets/scripts/views/location.js
+++ b/app/assets/scripts/views/location.js
@@ -240,30 +240,44 @@ var Location = React.createClass({
   },
 
   renderSourceInfo: function () {
+    const {data} = this.props.measurements;
+
     let sources = this.props.loc.data.sourceNames
       .map(o => _.find(this.props.sources, {name: o}))
       .filter(o => o);
 
+    if (data.attribution) {
+      // filtering attribution[0] b/c it kept showing up with same name and url as sources[0]
+      let added = data.attribution.filter((src, index) => index !== 0);
+      sources = [...sources, ...added];
+    }
+
     return (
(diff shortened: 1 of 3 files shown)


app/assets/scripts/views/location.js
```diff
@@ -240,30 +240,39 @@ var Location = React.createClass({
...
-                <div className='col-sec'>
-                  {source.description
-                    ? <p>{source.description}</p>
-                    : null}
-                  For more information contact <a href={`mailto:${source.contacts[0]}`}>{source.contacts[0]}</a>.
-                </div>
-              </li>)}
+              {sources.map(source =>
+                <li key={source.name}>
+                    <p><a href={source.sourceURL || source.url} title='View source information'>{source.name}</a></p>
```
jflasher: Not all attribution objects will have a URL. When this is the case, could you not use the <a> element so it doesn't show up as a link?

CamLatimer:
@jflasher I amended that last commit. Let me know if that needs to change.

jflasher:
Looks good, merging, thanks for the help!

PR #287, opened 2019-06-11
[WIP] Update API call for nearest locations

vgeorge:
The parameter `nearest` at `/locations` endpoint is deprecated. This PR changes the `GET /locations` request, using `order_by=distance&radius=99999999&limit=3`.

The big number on radius parameter was included to guarantee the query will return something, keeping the behavior of `nearest` param. There might be no locations if the default radius (2.5 km) is used and that is something the frontend is not design for.


vgeorge:
@jflasher please hold your review here until I fix Travis CI build failure.

vgeorge:
TravisCI error fixed, ready for review @jflasher. This is based on https://github.com/openaq/openaq.org/pull/288.

jflasher:
Thanks! Merging to staging for testing.

PR #288, opened 2019-06-11
Add .nvmrc and VSCode launch config

vgeorge:
I've tried to build the app with node 10.x (current LTS), but it doesn't work. So I added a .nvmrc file with the required node version and updated the install instructions to be more straightforward. 

Also added a VSCode launch config file to debug the app with Chrome. I think it's useful because the way the app is built currently doesn't match VSCode default config for Chrome debug.


vgeorge:
@jflasher please hold your review here until I fix Travis CI build failure.

vgeorge:
I'm puzzled with this TravisCI build error. It is related with gulp task `markdown`. This task interprets markdown files at `app/contents/projects`, transform to JSON and merge in `content.json` file. After running `nvm i` and `nvm i`, the following task runs ok and will generate a `content.json` file at `app/assets/contents`:

    ./node_modules/.bin/cross-env NODE_ENV=production ./node_modules/.bin/gulp markdown

But, when running `npm run build`, the same markdown task is executed as a sub-task. The file is not generated and an error is thrown. By adding a console.log at line 252, I see that the markdown task didn't found any input files.

When running `npm run serve` the same error is thrown, but a task re-run is triggered automatically by gulp-watch and the file is generated succesfully.

I've tried to replace gulp-util by list-stream, but the behavior is the same. The problem also happens in current dev branch. Any ideas?

vgeorge:
@jflasher The problem was solver, you can review this now.

jflasher:
Superseded by https://github.com/openaq/openaq.org/pull/287

PR #309, opened 2019-08-20
Update board info

olafveerman:
This updates the board info, and moves the markup into its own component. One that can be re-used for other teams in the near future, like the advisory board.


app/assets/scripts/views/about.js
```diff
@@ -138,47 +141,31 @@ var About = React.createClass({
...
-                <li>
-                  <article className='board-member'>
-                    <figure className='board-member__avatar'>
-                      <img src='/assets/graphics/content/board/board-avatar--placeholder.jpg' width='320' height='320' alt='Board avatar' />
-                    </figure>
-                    <h1 className='board-member__title'>Matthew Higgins</h1>
-                    <p className='board-member__role'>Treasurer</p>
-                  </article>
-                </li>
+              <TeamList items={_(content.governingBoard)
```
danielfdsilva: Since this is static content, you should sort outside the render function as to avoid creating new objects every time.

app/assets/scripts/views/about.js
```diff
@@ -138,47 +141,31 @@ var About = React.createClass({
...
+          <section className='fold' id='about-fold-board'>
+            <div className='inner'>
+              <header className='fold__header'>
+                <h1 className='fold__title'>Our advisory board</h1>
+                <div className='fold__teaser prose prose--responsive'>
+                  <p>Mauris posuere mauris a molestie ultrices. Donec risus ligula, rutrum laoreet elementum aliquam, ullamcorper in nibh.</p>
+                </div>
+              </header>
+
+              <TeamList items={_(content.advisoryBoard)
```
danielfdsilva: Same as above

PR #339, opened 2019-10-01
Update twitter links and meta

ddmck:
👋 I noticed the Twitter links and meta point to the old account.

I've updated them in:
app/index.html
app/assets/scripts/components/connect-fold.js
app/assets/scripts/components/page-footer.js


jflasher:
Thanks for catching this @ddmck!

RocketD0g:
Seriously! Thanks, @ddmck!

PR #343, opened 2019-10-05
FIX #304 Email not being passed to Mailchimp

### LINKED ISSUE #304: Email address when subscribing to newsletter
The 'subscribe to newsletter' feature asks people to provide their email.

image

Clicking the button leads to an empty Mailchimp form, and requires the person to fill in the email again, which is redundant.

Two options:

1. remove the email field from the website, and lead straight to the empty Mailchimp form
2. keep the field on the website, but populate the email field on the Mailchimp form with the value provided

ddmck:
Fixes: https://github.com/openaq/openaq.org/issues/304

The form reset on submit was stripping out the email value before making the post request.
I've removed the onSubmit function, but if you'd rather I implemented a timeout to remove it after the post, let me know.

Happy side-effect: I'm now on the mailing list as a result of testing 🎉 

Something to consider: Submitting this form subscribes without the need to add first and last name, if you want/need those details let me know and I can look into just pre-filling the form.

### DIFF
--- a/app/assets/scripts/components/page-footer.js
+++ b/app/assets/scripts/components/page-footer.js
@@ -9,6 +9,7 @@ var PageFooter = React.createClass({
   propTypes: {
     measurements: React.PropTypes.number
   },
+
   render: function () {
     let copyright = this.props.measurements !== null
       ? `${formatThousands(this.props.measurements)} measurements captured`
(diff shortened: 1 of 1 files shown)


jflasher:
@RocketD0g flagging for you as it affects newsletter emails/signups.

RocketD0g:
Thanks, @ddmck! It'd be great to have folks' first and last names too (though not absolutely needed on our end).

PR #347, opened 2019-11-02
Dockerfile for development and build

dolugen:


jflasher:
Thanks @dolugen! @sruti would you want to review this? Looks like it may help with initial startup when developing.

PR #372, opened 2020-03-06
Board updates (#371)

RocketD0g:
* Create robin.md

* Delete robin-wilson.md

* Update matthew-whitney.md

* Updated headshot of sarath


RocketD0g:
Moving Robin from advisory to gov board, updates Sarath's headshot.

PR #418, opened 2020-10-20
Feature/417 card list

jvntf:
Style results page as grid per mockups
image


rwegener2:
Looks good.  Once we are using the updated API it also might make sense to include adding the orange source tags, such as "LCS" or "mobile" in this PR.  That could also be saved for another PR.

AliceR (approved):

jvntf:
@rwegener2 added those tags here

PR #419, opened 2020-10-20
Feature/416 locations filters

jvntf:


jvntf:
@rwegener2 @alicer @tania-pires i'm not able to tag you for review directly (perhaps repo permissions @sruti ?) 

this PR reformats the locations hub filters as dropdowns, and adds a source filter. this is not yet connected to api.

rwegener2:
@jvntf Nice work here.

Two questions for the team:
* I am wondering about the labelling of the second filter. "Data Type" doesn't seem quite right.  I would have suggested "Measurement" or "Measurand".  I noticed the existing website uses "Value".  @sruti Do you have any insight into a good label for this?

* It looks like you can use multiple "order by" attributes, which I presume to be hierarchical (if there is a tie in the first filter then sort by the second filter).  Right now they show up next to the filters.  @tania-pires Do you have any low lift ideas for how to display the Order By selections that might make what is happening more obvious?

How the Filters (orange) and the Order By selections (blue) show up:
Screen Shot 2020-11-05 at 3 53 29 PM

I'll look forward to seeing what we do with the "Data Source" filter once this is hooked up to the API.

sruti:
> @rwegener2 @AliceR @tania-pires i'm not able to tag you for review directly (perhaps repo permissions @sruti ?)
> 
> this PR reformats the locations hub filters as dropdowns, and adds a source filter. this is not yet connected to api.

@jvntf I think I needed to add everyone as collaborators to be tagged. Invited them to the `lcs-pilot-frontend` team which should give them access. Let me know if this doesn't work or there's anyone else I need to add

rwegener2:
Update from the meeting: let's use "Pollutant" for the filter that right now currently says "Data Type"

jvntf:
Changed filter title. @rwegener2 @AliceR good to merge when ci (eventually) finishes?

PR #432, opened 2020-11-10
Upgrade to React 17

AliceR:
In an effort to upgrade to React 17, other dependencies need to be upgraded as well.
What I did:
1. Upgrade to React 17
2. Upgrade to React-DOM 17
3. Upgrade to PropTypes from separate package
4. Upgrade browserify/babelify/babel
5. Upgrade react-router to react-router-dom (incl. changes that need to be made to components)
6. ‘react-router-scroll’ not compatible with latest react router, removing for now
7. Patch React.createClass with drop-in replacement ‘create-react-class’ - this allows us to keep the old code without major refactoring
8. openaq-design-system relies on react 15 as peer dependency, needs to be upgraded as well: https://github.com/openaq/openaq-design-system/tree/feature/v3.1.0
9. openaq-design-system imports react-scrollbar that relies on react 15 as peer dependency. External lib that can not be easily upgraded, remove for now and replace with regular div
10. Replace ‘react-router-redux’ with ‘history’
11. Replace this.props.router with this.props.location
12. Connect Redux dev tools
13. Parse location.search string instead of using location.query object
14. Re-order routes top down, from most specific to least specific
14. Use matchPath to get pageClass prop from route
15. Use HashRouter instead of BrowserRouter
16. props.params doesn’t exist anymore, now in props.match

What I still want to look into:

- [x] Fix scroll boxes (after removing ScrollArea from openaq-design-system) -> overridden by @jvntf 's new filter dropdowns
- [x] Fix page transitions and scroll positions (after removing outdated react-router-scroll); [the docs](https://reactrouter.com/web/guides/scroll-restoration) might help
- [ ] Investigate console warning: 
>You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
- [x] Investigate failing CI

----------------------
⚠️We do not have tests in place, and I am not familiar with the project. I would encourage everyone to do a thorough manual QA on this branch before we release it to production. There might be unexpected issues with the upgraded libraries.


jvntf:
Resolved conflicts.

package.json
```diff
@@ -83,24 +83,25 @@
...
     "d3": "^4.2.1",
     "gulp-uglify-es": "^1.0.4",
+    "history": "^4.10.1",
     "isomorphic-fetch": "^2.2.1",
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.20",
     "mapbox-gl": "^0.50.0",
     "moment": "^2.14.1",
     "openaq-data-format": "github:openaq/openaq-data-format#master",
     "openaq-design-system": "github:openaq/openaq-design-system#feature/v3.0.0",
```
jvntf, line 95:
```suggestion
    "openaq-design-system": "github:openaq/openaq-design-system#feature/v3.1.0",
```

In order to get this to work I changed this version number, and then I removed `package-lock.json`, ran `npm install` and then had to manually install `npm install https://github.com/openaq/openaq-design-system\#feature/v3.1.0`. 

@AliceR thoughts? should my modified package-lock be committed now?

jvntf:
@AliceR this is so great thanks for your hard work getting this upgraded!!
See my above comment on getting it to use the correct version of `design-system`

package.json
AliceR: I forgot to commit this change, thanks! Added with 881f4a7

jvntf (approved):

AliceR:
Adding to the list of issues to look into:
- [ ] `/compare` is throwing an error, need to fix

PR #433, opened 2020-11-11
Feature/421 card

jvntf:


package.json
```diff
@@ -110,7 +110,6 @@
     "redux": "^3.5.2",
     "redux-logger": "^2.6.1",
     "redux-thunk": "^2.1.0",
-    "styled-components": "^5.2.1",
```
AliceR: ups, I think we need that!

app/assets/scripts/components/card-list.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { PropTypes as T } from 'prop-types';
+import styled, { css } from 'styled-components';
+
+const Wrapper = styled.div`
+  display: grid;
+  ${({gridGap, gridTemplateRows, gridTemplateColumns}) => css`
+    grid-template-columns: ${gridTemplateColumns || 'repeat(4, 1fr)'};
+    grid-template-rows: ${gridTemplateRows || 'auto'};
+    grid-gap: ${gridGap || '2rem'};
```
AliceR, line 10: :+1: I like this approach of default styles that can easily be overridden if needed.

app/assets/scripts/components/card.js
```diff
@@ -0,0 +1,106 @@
...
+import styled, { css } from 'styled-components';
+
+const CardWrapper = styled.article`
+  ${({gridColumn, gridRow}) => css`
+    grid-column: ${gridColumn || 'auto'};
+    grid-row: ${gridRow || 'auto'};
+  `}
+`;
+
+const CardContents = styled.a`
```
AliceR, line 12: Not sure if I understand the reasoning behind these empty styled-components?

app/assets/scripts/components/card.js
```diff
@@ -0,0 +1,106 @@
...
+    gridRow,
+    title,
+    subtitle,
+    renderHeader,
+    renderBody,
+    renderFooter
+  } = props;
+  return (
+    <CardWrapper
+      className='card' i
```
AliceR, line 59: the  `i` looks accidental?
```suggestion
      className='card'
```

AliceR (approved):
🙌 This looks great! I threw it into the new location view to test, and it works well. 
I have only a few minor comments, otherwise good to merge.

app/assets/scripts/components/card.js
jvntf: the reasoning was in case we need to override / augment any of the sass styles, i thought it's easier / more modern to do that in styled comps? also for readability?

can also see the argument that it's making things more complicated to introduce styled component styling on top of the sass. what do you think?

AliceR: Migrating to styled-components makes sense to me, in general. I meant to question only the empty ones - however I am fine with leaving them as is.

PR #442, opened 2020-11-17
Feature/420 dashboard table

jvntf:
Compared on dashboard cards to show changes (this is branched from `feature/436-dashboard-cards`)


app/assets/scripts/views/location/averages-card.js
```diff
@@ -0,0 +1,117 @@
...
+import Table from '../../components/table';
+import { shortenLargeNumber } from '../../utils/format';
+
+const initData = {
+  pollutant: {
+    values: [],
+    formatHeader: v => v.toUpperCase(),
+    style: {
+      color: 'black',
+      fontWeight: 700,
```
necoline, line 15:
```suggestion
      fontWeight: 700,
      textAlign: 'center',
```

app/assets/scripts/views/location/averages-card.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+const StyledLoading = styled(LoadingMessage)`
+  grid-column: 1 / 7;
+`;
+const ErrorMessage = styled.div`
+  grid-column: 1 / 7;
+`;
+
+// TODO This function currently just extracts the first data available for each pollutant
```
necoline, line 43: What is the task to complete here?

app/assets/scripts/components/table.js
```diff
@@ -0,0 +1,164 @@
...
+ *        formatCell: [function](optional),
+ *        formatHeader: [function])(optional),
+ *        sortable: [boolean](optional, default false),
+ *        cellWidth: [string](optional)
+ *      }
+ *    }
+ */
+const Table = styled.table``;
+const Row = styled.tr`
+  background-color: ${({ index }) => (index % 2 === 0 ? '#ffffff' : '#f4f9fe')};
```
necoline, line 29: Should we be referencing a theme for these colors so we don't have to hunt them down in later component development?

app/assets/scripts/actions/average-measurements.js
```diff
@@ -0,0 +1,73 @@
...
+              data.results = data.results.concat(json.results);
+            }
+            if (page * limit < json.meta.found) {
+              return fetcher(++page);
+            } else {
+              return dispatch(receiveAverageMeasurements(data));
+            }
+          },
+          e => {
+            console.log('e', e);
```
necoline, line 63:
```suggestion
            console.error('Error:', e);
```

necoline (approved):
Added a few comments, but looks good!

app/assets/scripts/views/location/averages-card.js
jvntf: We need to confirm what data this table should display. The endpoint returns either all available average data, or a selected date range. For now this just uses the most recent result (also the version of the api that is currently connected is only available for a single pollutant. cc @rwegener2

app/assets/scripts/components/table.js
jvntf: yes i agree that using explicit colors here is not the best choice. what would be ideal is if we can set up a theme to parse out variables from the sass files. i think lightest fix for this is to use class name and set styles with sass variables. what do you think? @necoline @AliceR

app/assets/scripts/components/table.js
```diff
@@ -0,0 +1,164 @@
...
+          {Object.entries(data).map(([key, datum]) => {
+            const { formatHeader } = datum;
+            return (
+              <Header
+                className={`sort_${sortAsc ? 'asc' : 'desc'}`}
+                key={key}
+                active={key === activeHeader}
+                onClick={() => {
+                  setActiveHeader(key);
+                  setSortAsc(!sortAsc);
```
AliceR, line 109: Looks like you'd actually want to split the state to control the columns sort individually?

AliceR (approved):

app/assets/scripts/components/table.js
jvntf: What do you think the functionality should be? I've implemented as follows but not sure it is most intuitive

click on a header -> toggles sorting direction and sets sorting to be based on that column

cc @tania-pires @AliceR

app/assets/scripts/views/location/averages-card.js
necoline: I see, can we add a bit of this context to the comment so it's easier for us to pick up later?

PR #444, opened 2020-11-19
Projects hub page

AliceR:
Add projects hub page, as well as filter and card components. Not yet populated with data.
Do we have a mock of the expected projects data available? Some api definition I can use as baseline?

Wonder what the best next step is. In the long run, I think it would make sense to extract the filter component, as well well as the card component, to be reusable across hubs (projects, locations, countries). What do you think, @jvntf?
(Note: so far I only moved around the existing code, the component composition could certainly be improved)


jvntf:
@AliceR looking good, yes i agree that it would be good to abstract the cards and filters. I believe that you can request countries rather than 'projects' for demo data. I think the data is basically the same just aggregated differently? cc @bitner

jvntf:
@AliceR on clicking `view more`, this breaks, is that expected / unimplimented? otherwise looks good

AliceR:
We don't have a `projects page` to link to, yet. Would be better to disable the link for now though instead of breaking the app. I'll have a look.

AliceR:
[Done](3dea77d)! Could you review again please, @jvntf ?

jvntf (approved):

PR #445, opened 2020-11-19
Feature/420 bar chart2

necoline:
I built this on @jvntf 's scatter-plot PR since there was a lot that of work already done. 
I started out with a d3 chart, but it was taking too much time to get it all looking right, so I opted to use chart.js for an easy implementation. 

I realize that I'm not bringing in all of the data. It's just the data from the first page, I thought I would just get this up for now and make that change later. The code should support new dumps of data to create an aggregate data set for the chart.


app/assets/scripts/components/bar-chart-measurment.js
```diff
@@ -0,0 +1,37 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { Bar } from 'react-chartjs-2';
+
+export default function BarChart({ title, frequency, xAxisLabels }) {
+  const data = {
+    labels: xAxisLabels,
+    datasets: [
+      {
+        label: 'measurments',
```
AliceR, line 10:
```suggestion
        label: 'Measurements',
```

app/assets/scripts/components/bar-chart-measurment.js
```diff
@@ -0,0 +1,37 @@
...
+        data: frequency,
+        backgroundColor: '#198CFF',
+      },
+    ],
+  };
+
+  const options = {
+    scales: {
```
AliceR, line 18: I think we can hide the legend, like this:
```suggestion
    legend: {
      display: false,
    },
    scales: {
```

app/assets/scripts/components/bar-chart-measurment.js
```diff
@@ -0,0 +1,37 @@
...
+      yAxes: [
+        {
+          ticks: {
+            beginAtZero: true,
+          },
+        },
+      ],
```
AliceR, line 25: We can disable the vertical grid lines:
```suggestion
      ],
      xAxes: [
        {
          gridLines: { display: false },
        },
      ],
```

app/assets/scripts/views/location/temporal-measurements.js
```diff
@@ -0,0 +1,213 @@
...
+            }}
+          />
+          <CardTitle>Temporal Coverage</CardTitle>
+        </CardHeader>
+      )}
+      renderBody={() => (
+        <div className="card__body">
+          {activeTabMeasurments.length ? (
+            <CardList gridTemplateColumns={'repeat(3, 1fr)'} className="inner">
+              <Card
```
AliceR, line 145: Should each chart be a Card within the card? Seems odd to me... I'd think the Card component has to many features we don't need here at all. Just a div could do?

app/assets/scripts/views/location/location.js
```diff
@@ -157,6 +158,13 @@ function Location(props) {
             activeParam={getActiveParameterData()}
             onFilterSelect={onFilterSelect}
           />
+
+          <TemporalMeasurements
+            measurements={props.measurements}
+            parameters={props.parameters}
+            activeParam={getActiveParameterData()}
+            onFilterSelect={onFilterSelect}
+          />
```
jvntf, line 167:
```suggestion
            measurements={props.measurements}
            parameters={props.parameters}
          />
```

app/assets/scripts/views/location/temporal-measurements.js
```diff
@@ -0,0 +1,213 @@
...
+    monthOfYear.forEach(month =>
+      monthCount[month]
+        ? (sortedMonthCount[month] = monthCount[month])
+        : (sortedMonthCount[month] = 0)
+    );
+    return sortedMonthCount;
+  };
+
+  if (fetching) {
+    return <LoadingMessage />;
```
jvntf, line 108: Id recommend wrapping this in a styled comp, see measurements-card.js for example. otherwise the loading comp is not the same size as the card

app/assets/scripts/views/location/temporal-measurements.js
jvntf: I agree

app/assets/scripts/views/location/temporal-measurements.js
```diff
@@ -0,0 +1,213 @@
...
+                    />
+                  </div>
+                )}
+                renderFooter={() => null}
+              />
+              <Card
+                renderBody={() => (
+                  <div className="card__body">
+                    <TemporalChart
+                      title="Month of the Year"
```
jvntf, line 173: Are these sections of the data guaranteed to exist?

jvntf (changes requested):
some tiny changes but looks great!

app/assets/scripts/views/location/temporal-measurements.js
AliceR: Hm, I think they should - isn't it just a different temporal breakdown of the same data always?

AliceR:
@jvntf I addressed the changes, could you review again, please?

app/assets/scripts/components/bar-chart-measurement.js
```diff
@@ -0,0 +1,53 @@
...
+    defaultFontFamily:
+      "'Roboto', 'Helvetica Neue', Helvetica, Arial, sans-serif",
+    legend: {
+      display: false,
+    },
+    scales: {
+      yAxes: [
+        {
+          ticks: {
+            beginAtZero: true,
```
jvntf, line 27:
```suggestion
            beginAtZero: true,
            maxTicksLimit: 5
```

I suggest adding this limit, too many horizontal lines looks messy

jvntf (approved):
@AliceR nice! approved, one suggestion above

app/assets/scripts/components/bar-chart-measurement.js
AliceR: Great! I couldn't figure out how to do it, thanks!

PR #446, opened 2020-11-23
Feature/423 scatter plot

jvntf:


jvntf:
This contains bar chart and scatter plot changes

AliceR (changes requested):
Looks good so far. Only the time picker is not doing what I'd expect: it sets the end date to `null` when I select the first date.

AliceR:
I am going to add a new ticket for the bug and merge this PR into the lcs branch. I need those components for further development.

PR #450, opened 2020-11-26
Add project page and some dashboard components

### LINKED ISSUE #427: Project dashboard
Setup generic layout to display cards in grid 
image

AliceR:
Closes #427


app/assets/scripts/components/bar-chart.js
```diff
@@ -2,13 +2,15 @@ import React from 'react';
...
+import { round } from '../utils/format';
+
+export default function BarChart({ data, xAxisLabels }) {
+  const series = {
     labels: xAxisLabels,
     datasets: [
       {
-        label: 'Measurements',
-        data: frequency,
+        label: 'average',
```
jvntf, line 12: What is this label used for? just wondering if it should be passed as props or ok to hardcode?

jvntf (approved):
This looks great! 
I noticed a bug while reviewing this, i don't think its related so approving

One is with the Routing -> navigating `back` and `forward` seems to break pages. Do you see the same @AliceR ?

jvntf:
Also seeing a lot of errors produced on locations pages. im assuming these have to do with API updates / we need to make updates there on separate PR? looks like issue with API response format @AliceR

AliceR:
@jvntf indeed, fixed here: 08b0ffc

app/assets/scripts/components/bar-chart.js
AliceR: it appears on the legend (currently hidden) and the tooltips. Passing as props is a good idea in this case, as it could just as well be the count (see [this line](https://github.com/openaq/openaq.org/pull/450/files/0267fba245319314df603234823dcaaf7753dd20#diff-afda5e67a8c150e5d2106efb699b7eb2f95fdc0c7d8fca99f321a724fd3005b6R248))

AliceR:
There are a lot of bugs, still. I'll merge this into lcs, so we can tackle them together from there. Adding a card for the routing bug [on the project board](https://github.com/openaq/openaq.org/projects/1#card-50319197).

PR #455, opened 2020-12-01
Update/locations hub

jvntf:
Refactor locations hub


app/assets/scripts/views/locations-hub/location-card.js
```diff
@@ -0,0 +1,101 @@
...
+  countryData,
+  sourcesData,
+}) {
+  let updated = moment(lastUpdated).fromNow();
+  let started = moment(firstUpdated).format('YYYY/MM/DD');
+
+  const country = countryData || {};
+  let sources = [];
+  if (sourcesData.length) {
+    sourcesData.forEach((o, i) => {
```
AliceR, line 28: I think we can ask for this information to be included in the api response, rather than matching it across fetched data in the frontend (cc @bitner )

We would need per `/locations` location object:
- list with sources that contain url and name
- the country name (currently only has country code)

AliceR (approved):
👍 

If you'd want to spend some more time refactoring, I think we can remove both projects and locations from redux and use the internal state on those hub pages. We really don't need this data anywhere else.

AliceR (changes requested):
we need to check this line to fix the filter: https://github.com/openaq/openaq.org/pull/455/files#diff-3c6201c6b67ad66384e2613f62084054eefb60abe475eab4976039b275026f37R29

I just learned that the request should be `...country=TW&country=RO` instead of `country=TW%2CRO`

Though, to be fair, is not related to this change and was likely broken before 😬

app/assets/scripts/views/locations-hub/location-card.js
bitner: is this for the locations/<id> endpoinst or the locations?location= endpoint?

AliceR: For the list of `/locations` used here https://openaq-staging.org/#/locations

Current response format 
```json
[{"id":2,"country":"GH","city":"Accra","cities":["Accra"],"location":"JTT - Jamestown","locations":["JTT - Jamestown"],"soureName":"Dr. Raphael E. Arku and Colleagues","sourceNames":["Dr. Raphael E. Arku and Colleagues"],"sourceType":"government","sourceTypes":["government"],"coordinates":{"longitude":-0.2129917,"latitude":5.5334556},"firstUpdated":"2006-11-13T07:31:00+00:00","lastUpdated":"2007-07-11T07:25:00+00:00","parameters":["pm25","pm10"],"countsByMeasurement":[{"parameter":"pm25","count":32},{"parameter":"pm10","count":32}],"count":64}, ... ]
```

jvntf:
@AliceR thanks i think better to do that refactor in separate PR, can also merge locations and projects to reuse most of the same code there

AliceR (approved):

app/assets/scripts/actions/locations.js
```diff
@@ -27,6 +27,7 @@ export function fetchLocations(page = 1, filters, limit = 15) {
     dispatch(requestLocations());
 
     let f = buildAPIQS(filters, { arrayFormat: 'repeat' });
+    console.log(filters)
```
AliceR, line 30:
```suggestion
```

PR #463, opened 2020-12-08
Add basic integration test for each page

AliceR:
This adds very basic integration test for each page. It only checks for presence of 'header', 'main', 'footer', and 'h1' elements, which does not say much yet about the actual page content (especially because every page initially shows a global loading that contains all these elements). The about.spec.js already has some more checks, as well as navigation.spec.js testing the main navigation. More tests to follow.

Further I added a github action workflow to run lint and integration tests on push and pull request events.


.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,71 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Use Node.js 10
+        uses: actions/setup-node@v1
+        with:
+          node-version: 10
+
+      - name: Cache node modules
```
necoline, line 21: great job sorting out the caching!

necoline (commented):
This looks like a great start. How are you testing this locally? It would be great to get a script using something `start-server-and-test` up and running.

AliceR:
Thanks @necoline , I added a guide here: https://github.com/openaq/openaq.org/tree/integration-test/cypress - you'd need to start the dev server and then `npm run cy:open`.

We could add additional scripts to run tests with its own dev server instead of requiring to start the server manually, but it comes with cost of more information to write and read, more potential for confusion, and an additional dependency ([`start-server-and-test`](https://github.com/bahmutov/start-server-and-test)). What do you think, should we add it?

necoline:
Ah ok, It turns out the issue I was facing, was separate to the issue of running a test server (documented [here](https://github.com/cypress-io/cypress/issues/2610))

I wonder if we can get a testing server running out of gulp? I'll poke around on this, but I don't think this is worth spending much time on.

AliceR:
I rebased this branch onto api-v2 to get the api changes and fixes.

AliceR:
Finally getting closer to running the tests on ci 😅 
I added a new config file as well as a script to explicitly specify the api that should be used by the tests.
The two failing tests now are a small type in the location hub ([fixed on develop](https://github.com/openaq/openaq.org/commit/1743fd87b9eae90268ab65552e4c0f260991fe81)) and a server error on the projects endpoint.

jvntf:
confirming what @AliceR says above. 
How is it that the typo change is fixed on develop but not here? should we merge develop to this branch?

jvntf (approved):

PR #464, opened 2020-12-09
Use v2 for locations hub

jvntf:
Update locations hub to v2. 
After this is merged, locations single page can be pull location data from props rather than requesting again. cc @necoline

### DIFF
--- a/app/assets/scripts/config/staging.js
+++ b/app/assets/scripts/config/staging.js
@@ -6,5 +6,6 @@
 module.exports = {
   environment: 'staging',
   api: 'https://ytr9800fbk.execute-api.us-east-1.amazonaws.com',
+  apiv2: 'https://ytr9800fbk.execute-api.us-east-1.amazonaws.com/v2',
   metadata: 'https://metadata.openaq-staging.org',
 };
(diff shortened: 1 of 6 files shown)


app/assets/scripts/config/staging.js
```diff
@@ -6,5 +6,6 @@
 module.exports = {
   environment: 'staging',
   api: 'https://ytr9800fbk.execute-api.us-east-1.amazonaws.com',
+  apiv2: 'https://ytr9800fbk.execute-api.us-east-1.amazonaws.com/v2',
```
AliceR, line 9: I think we should be able to simply add /v2 to the above url and use the v2 api for all requests. That way we don't have to maintain two versions, making us wonder when to use which one.

necoline (commented):
I don't know if this is the case for all or just some of the `sources` on the `SourceCard` but I ran into errors with some sources coming back as an object rather than an array. I don't know if we want to fix this here or on the backend, but I added a fix in a [PR ](#465 )

necoline:
Also should this be pointing to `lcs` rather than `develop`?

AliceR:
I merged the changes made here into #468

PR #465, opened 2020-12-09
Fix/462 source type

necoline:


AliceR:
In which cases is the sources not an array? I wonder if it would be better to ensure we always pass in an array in the expected format to the component and fix this issue higher up in the data flow. Potentially at the api?

necoline:
I haven't found a case that it's _not_ an object and so it was a necessary adjustment for me to see the page without anything breaking. It would be helpful for me to get clarity on why this doesn't seem to come up for you both before we can identify the edge-case here.

AliceR:
We asked the backend to provide the source as list, to be consistent with other response objects. It should make this change unnecessary.

PR #466, opened 2020-12-09
Fix/locations address

necoline:
- [x] refactors countries-hub
- [x] fixes data changes coming from new location endpoint (location id and params)
- [x] adds some fallbacks


AliceR (approved):
Thanks for refactoring the Countries Hub 🙌 I am glad only small additions were necessary in order to reuse the header and card components.

AliceR:
Before merging: Note that this branch contains commits from #465 with pending conversation around sources.

PR #468, opened 2020-12-10
Update to api v2

AliceR:
@jvntf as soon as @bitner has his changes up, we should be able to use the new api as complete replacement.
This simplifies a lot of things in the components we already refactored, but also requires a few more changes in other places - mainly, the home page: we would need to refactor the current implementation of the location comparison cards.


app/assets/scripts/views/locations-hub/results.js
```diff
@@ -85,13 +76,13 @@ export default function Results({
               id={loc.id}
               onDownloadClick={openModal}
               key={loc.location}
```
AliceR, line 78:
```suggestion
              key={loc.id}
```

jvntf:
@AliceR I'm running into some issues testing this. I'm wondering if it is config related? I see these requests failing.

I copied the staging config into my `config/local`, so these requests are going to v2. is that correct?
image

AliceR:
ah, it seems @bitner has not yet updated the api, he gave me a "develop" api to run this with. I'll send it to you so you can test it.

jvntf (approved):

jvntf:
Everything looks good, just noting that i was not able to test projects hub / projects page due to pending updates to api @AliceR cc @bitner

AliceR:
Merging to lcs to use as base for further development. We should test those pages again once backend is back up 👍

PR #469, opened 2020-12-10
adds alphabetical sort to filters

necoline:
adds sort to location hub and dataset page

### DIFF
--- a/app/assets/scripts/views/locations-hub/filter.js
+++ b/app/assets/scripts/views/locations-hub/filter.js
@@ -34,6 +34,13 @@ export default function Filter({ countries, parameters, sources }) {
     initFromLocation(qs.parse(location.search, { ignoreQueryPrefix: true }))
   );
 
+  // alphabetizes filter names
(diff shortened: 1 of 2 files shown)


AliceR:
We need to send the sort order with the api request, otherwise only whats on the paginated result (the current 15 items) will be sorted. In https://github.com/openaq/openaq.org/pull/468/commits/3d6addaf844c240d4737b59554b719b8d8db53b8 I added the sort key to the filters object, which will be sent as query parameters to the api. I'm sorry, I should have provided a better issue description on the card (and also, made it clearer this was implemented in #468). 
🙈(sorry)

AliceR (approved):
For reference - mostly for myself, as I was blind earlier - this is the card we talk about here https://github.com/openaq/openaq.org/projects/1#card-50715872

(not [this one](https://github.com/openaq/openaq.org/projects/1#card-50415933))

necoline:
I think I made things difficult by adding an ambiguous PR name. Maybe tickets are worth it for the small issues for the linking 🤷

PR #471, opened 2020-12-11
Lcs

jvntf:


jvntf:
This should be good to merge, but depends on having this api in your local config:  'https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2'. Projects hub is not returned by this api yet.

Marking as draft until all ready, then will add this api string to the staging config. and to production? 
@bitner @AliceR

AliceR (approved):

PR #474, opened 2020-12-14
Add mobile tag

jvntf:


app/assets/scripts/components/card/card.js
```diff
@@ -72,11 +78,14 @@ export default function Card({
...
               {title && <CardTitle className="card__title">{title}</CardTitle>}
             </CardHeadline>
-
-            {tags && typeof tags === 'string' ? (
-              <CardTag label={tags} />
-            ) : (
-              tags && tags.map((tag, i) => <CardTag key={i} label={tag} />)
+            {tags && (
+              <TagWrapper>
+                {typeof tags === 'string' ? (
```
AliceR, line 83: Irrelevant detail, but I wonder if it would be easier to always pass in an array, even when we only have 1 tag in the list. Maybe not easier but likely prettier :)

AliceR (approved):

app/assets/scripts/components/card/card.js
jvntf: added https://github.com/openaq/openaq.org/projects/1#card-51238126

PR #475, opened 2020-12-14
Fix/redundant filters

necoline:
I spoke with Bitner about about how reduce the redundancy in the parameters list from something like this:

```
{id: "co2", name: "CO2", description: "CO2", preferredUnit: "ppm"}
{id: "co2", name: "CO2", description: "CO2", preferredUnit: "umol/mol"}
```
to this 
``` {id: "co2", name: "CO2", description: "CO2", preferredUnits: ["ppm", "umol/mol"]}```

Bitner mentioned that the client would be resistant to this type of modification, but that he could add an id for a measurand so we have something better to filter by. It doesn't really help anything in the case of the duplicate dropdown filters

I opted to create a reference a list of unique parameter ids in the dropdown list that can be consumed by the `filterSelect`

I also added some tests to check spot check for duplicates in the list, I also moved our cy gets from classnames and elements which tend to be dynamic to be cypress data selectors 

ref:
image


AliceR:
Cool, thanks for checkin in on possible backend solutions. I guess it's not too bad to parse this on the frontend. Unfortunately I still seem to get a few duplicates (SO2, NO2, CO, ... even BC, your test should fail at that one) - am I looking at the wrong filter list? The screenshot is from locations hub.

Screenshot 2020-12-14 at 21 30 28

AliceR:
Not sure if it relates to this change, but there seems to be an issue with unselecting chips as well

necoline:
strange, I don't know why we are seeing different results here, let's sync up tomorrow

AliceR:
[contains()](https://docs.cypress.io/api/commands/contains.html#Usage) yields only the first element, [find()](https://docs.cypress.io/api/commands/find.html#Selector) yields all that match. Added a few more data-cy selectors with the label in order to access items by name, didn't figure out how to query without that?

necoline:
Thanks @AliceR I love the approach of fixing the tests to fail on the presenting problem.
I was pointing to the wrong endpoint which had duplicate ids and names the recent change to having unique ids is preferable, but the set needed to updated to point to names rather than ids in order to filter out the duplicates.

necoline:
I'm not sure why the wait was added to the tests but I added it in when I merged the `lcs` only to get lint errors. I removed them per [this](https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting) guidance.

AliceR (approved):

PR #477, opened 2020-12-14
Pass param name instead of id

jvntf:
API update bugs / regressions

- [*]  fixed locations hub parameter filter sending id instead of name
- [*] fixed projecrts hub param filter sending id instead of name


AliceR:
Removing the filter pill throws an error, see failing test in e6ccdcc - seems like we are not using name vs id consistently?

jvntf:
@AliceR thanks for catch, fixed

AliceR (approved):

PR #482, opened 2020-12-15
Feature/481 date selection

### LINKED ISSUE #481: Choose a unique date range for the time series plots (locations & datasets dashboards)
## Goal
Allow the user to select a custom date window for data aggregations on each of the dashboards (Locations and Datasets), but limit them to aggregations that the backend can do quickly.

## Details
The user flow is shown in the first two screens of [this set of wireframes](https://projects.invisionapp.com/share/4PZP8PX96ZU#/screens/440256751).  The node selection feature is also included in that link, but this ticket just refers to date selection.

### User Flow
The flow of the date selection is:
1. A user arrives at the dashboard with the default selection of "entire lifetime of project" ([screen](https://projects.invisionapp.com/share/4PZP8PX96ZU#/screens/440256751)). All data on the page is aggregated over the lifetime of the project (more details below)
2. The user could instead choose to select a specific time window.  If they do that a second date selection tool pops up.  ([screen](https://projects.invisionapp.com/share/4PZP8PX96ZU#/screens/440256752)). This tool allows the user to select either a single day or a single month at a time to aggregate over.  We aren't allowing any additional flexibility due to backend performance.

### Dashboard Data Display
When we change the date on top the following components should change to reflect the updated time window:
* Time Series - show a time series for the given date range, binning by a different value depending on the size of the window _one day of data_ - hourly binning _one month of data_ - hourly binning (maybe daily instead?) _lifetime of project_ - daily binning?  We probably shouldn't show more than a year's worth of data at a daily resolution.  I think we should either never show more than a year's worth of data or we always show the full project lifetime but keep it at monthly resolution.  I'd lean toward the first one.
* Measurands - show the average value and total count for each pollutant over the time window (single day, single month, or project lifetime)
* Temporal Coverage - show the binned results for temporal coverage over the time window, adjusting in response to each bin.  _Hour of day_ - ignore for now, we are dropping this for all stationary locations.  _Day of week_ - visible for entire project lifetime and a single month, greyed out for single day _month of year_ - visible for entire project lifetime, greyed out for single day or single month

**Technical decision note** - the above descriptions assume that the lifetime of a sensor is at least a few months long.  If the sensor has only been in operation for a month the data display may not make sense.  I'm not sure what the best approach is to deal with this.  If we need to adjust how we are displaying or what we are displaying that is fine with me.  Feel free to ping me to brainstorm.

jvntf:
closes #481


jvntf:
@rwegener2 can you give this a functionality review please? hoping i captured the ticket correctly. I havent changed any of the logic for data rendering in the temporal coverage chart.  The API is returning a single day of data when I request DOW temporal resolution with one month of time specified. again not sure if this is correct behavior.

necoline (approved):
looks good! you have a few stray console logs and proptype errors but otherwise looks good to go.

necoline:
The date-picker window seems constrained by the dropdown width
image

jvntf:
@necoline how can i replicate? i dont see the same 
image

necoline:
🤔 I'm not sure, I'd love to get someone else to try this out.

AliceR:
The calendar seems to be showing up correctly for me. 
but I thing the whole filters should move into the page center, where we have the other content? Add className="inner", maybe?
Screenshot 2020-12-18 at 17 54 01

jvntf:
I added the date selector into the inner and added it to the project page. Was not able to test this because im seeing the request for base data fail. this is a small change since last test so hopefully this is still working. cc @necoline @AliceR

AliceR (changes requested):
I am running into a bunch of errors here. Can't suggest directly in this PR as there are no nearby changes.

in the parameters list, the `measurand` is named `parameter` (relates to #491, do we need to change the api response?):
```
+++ b/app/assets/scripts/views/projects-hub/project-card.js
@@ -48,7 +48,7 @@ export default function ProjectCard({
             {
               label: 'Parameters',
               value: parametersList
-                .map(p => p.measurand.toUpperCase())
+                .map(p => p.parameter.toUpperCase())
```

In the project view, it seems you forgot to access history from props:
```
+++ b/app/assets/scripts/views/project/project.js
@@ -30,8 +30,8 @@ const Dashboard = styled(CardList)`
   padding: 2rem 4rem;
 `;
 
-function Project(props) {
-  const { id } = props.match.params;
+function Project({ match, history, location }) {
+  const { id } = match.params;
 ```

Lots of proptypes error on the Location page, and the page fails to load. Do we need to adjust to api changes?

necoline (approved):

necoline:
I added a condition to the error message so if there is no data to display on a specific date, users are instructed to try a different date, I also fixed the styles and made some adjustments to respond to the api changes

PR #484, opened 2020-12-17
Refactor/country

necoline:
- [x] updates components to hooks
- [x] makes requests for country and locations in the component and removes redux actions
- [x] adds tests
- [x] adds header stats to common component
- [x] adds a few other common components
- [ ] moves the map to separate component - started but not complete


app/assets/scripts/views/country/country.js
```diff
@@ -0,0 +1,279 @@
...
+    };
+
+    const fetchCountry = id => {
+      setCountry(state => ({
+        ...state,
+        countryFetching: true,
+        countryError: null,
+      }));
+      let limit = 1;
+      fetch(`${config.api}/countries?limit=${limit}&country=${id}`)
```
AliceR, line 84: Maybe we should change the /countries endpoint to accept `​/countries​/{country_id}`, for consistency with /locations and /projects. What do you think, @bitner ?

AliceR:
Almost done, this looks very good already. Don't worry too much about the map component, it should be tackled in #476 .
Just the lint and a little bit of cleanup (old proptypes, unused selector) and this is good to merge.

app/assets/scripts/views/country/country.js
```diff
@@ -0,0 +1,279 @@
...
+      ).map(JSON.parse)
+    : null;
+
+  let locationGroups = _(locations).sortBy('city').groupBy('city').value();
+
+  return (
+    <>
+      {(countryError || !country) && <ErrorHeader />}
+      {countryFetching && <LoadingHeader />}
+      {country && locations && (
```
AliceR, line 139: should we use some if/else ?/: to ensure error, loading, and actual countries page are mutually exclusive?

necoline: ah yes, that is better

PR #486, opened 2020-12-17
Feature/473 mobile filter

jvntf:
Add isMobile filter
https://github.com/openaq/openaq.org/projects/1#card-50583341

Note at time of posting API returns 0 results when mobile flag is true OR false

cc @bitner

### DIFF
--- a/app/assets/scripts/views/locations-hub/filter.js
+++ b/app/assets/scripts/views/locations-hub/filter.js
@@ -75,6 +75,23 @@ export default function Filter({ countries, parameters, sources }) {
         break;
       }
 
+      case 'source_type': {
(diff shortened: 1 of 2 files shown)


necoline (commented):
I'm not sure why, but any selection on the dropdown will set `mobile: true`

image

jvntf:
@necoline can you confirm what the value isMobile is in the request? I think this may be an api issue. On my end im seeing isMobile correctly set to true / false in the GET request

necoline:
you're right, the filter param is alternating correctly. I was/am turned around by the fact that the initial setting of the filter seems to work, but it looks like you are working that out with @bitner

necoline (approved):

PR #487, opened 2020-12-17
fixes names and adds tests

necoline:
Standardizes naming: change all occurrences of "value" "pollutant" or "measurand" to "Parameter" -- locations filter; locations hub cards; dataset "measurands" table

adds tests


jvntf (changes requested):
Measurands card still uses 'pollutant', should that param? other wise looks good

necoline:
Thanks for catching that!

PR #489, opened 2020-12-17
Update/473 mobile pages

### LINKED ISSUE #473: Distinguish between mobile and stationary Locations Dashboards
## Smaller differences
The majority of the contents of the stationary and mobile dashboard pages should be the same.  Initially the distinction to make will be to add a label on the top of each type of page labelling the location as "mobile" or "stationary", as in this wireframe screenshot:

Screen Shot 2020-12-13 at 9 11 58 PM

The only other difference is that the "Temporal Coverage" component only needs to have "day of week" and "month of year", while the mobile version should retain the "hour of day" section.

## Nearby Locations
Once Nearby Locations is added to the locations page, the mobile maps might have to be adjusted slightly because they will be rendering a string of points instead of just a single data point.  Nearby locations may also show up which are also mobile nodes, so we will also have to visualize those as a unit.

The person picking up this task will have a bit of creative thinking to do about how to visualize the single mobile location and how to visualize many mobile locations, in part informed by the output of [this card](https://github.com/openaq/openaq.org/projects/1#card-50394816) which should reintroduce the Nearby Locations component to the dashboard page.

jvntf:
closes #473 

cc @rwegener2

### DIFF
--- a/app/assets/scripts/components/header/header.js
+++ b/app/assets/scripts/components/header/header.js
@@ -2,6 +2,15 @@ import React from 'react';
 import { PropTypes as T } from 'prop-types';
 import { Link } from 'react-router-dom';
 
+import styled from 'styled-components';
+
+const Tags = styled.div`
+  display: grid;
+  grid-gap: 0.5rem;
+  width: fit-content;
+  grid-template-columns: 1fr 1fr;
+`;
+
 export default function Header({
   id,
   tagline,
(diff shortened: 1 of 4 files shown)


necoline (commented):
The implementation looks right, but I'm not getting the `.header__pill` styles to show up

jvntf:
what are you seeing @necoline ?

jvntf:
some times i have issues with gulp not rebuilding after branch change until i save a file or something

necoline:
Got it! Rebuilding is always a strange ordeal, but it's working for me now

necoline (approved):

PR #492, opened 2020-12-18
Fix/number rounding

necoline:
Rounds all pollutant measurements on the dashboard pages to the nearest whole number - 
- [x] Latest measurements
- [x] Measurands table and adjust styles
- [x] Time Series hover


AliceR:
I know this was what the ticket said should be done, but...
Screenshot 2020-12-18 at 19 54 08
Are we sure about this change? It doesn't look good on the charts
Screenshot 2020-12-18 at 19 55 26
and latest measurements aren't quite useful anymore either being just 0.

rwegener2:
Thanks for noticing @AliceR.

EDF flagged this in a recent email - it sounds like the "round to the nearest whole number" was said by folks thinking about the values in ppb (Ex. 40 ppb).  Any values reported in ppm would show up instead as a decimal (Ex. 0.04 ppm).  @bitner will work on converting these to a more standard unit at some point.  We are waiting on some follow up from EDF.  So long term we will work on more standard units, and we can also do something like "<1 ppm" for numbers that still come back rounded to zero.

In the meantime let's round everything to the nearest hundredth (Ex. 9.99).

rwegener2:
@AliceR What is happening with Day of the Week?  Is that because of the rounding?

AliceR:
@rwegener2 yes, the day of week charts shows values to be either 0, 1 or 2 - no nuances in between - because of the rounding.

rwegener2:
@AliceR The number rounding actually doesn't concern me too much because that figure really shouldn't be showing pollutant measurements anyway, it should be showing measurement counts.  I'm fine to ignore that for now, or, if it's easy, just switch to counts.

What I was wondering about was actually the x-axis, labelled "Sat, Fri, Thur, Wed, Wed, Wed, Tue, Tue, Mon".  Is that an API issue?  Maybe connected to the situation where the pollutant with different units shows up multiple times?

necoline:
It looks like the data coming through on the DOW week data is coming through slightly differently. I think this would be best handled in a separate PR - I started working on this but will likely push a change after our deploy today

PR #514, opened 2021-01-05
Add/495 info button

### LINKED ISSUE #495: Dashboards - Add info hover text for dynamic components
## Overview
The goal is to clarify for the user the scientific details of what is happening in the more complex dashboard components.  This is especially important because the components that change dynamically with selected nodes or date ranges.

A small info button should be placed next to the title of each component, a shown in [these wireframes](https://projects.invisionapp.com/d/#/console/20815156/437956789/preview).  The text below would show up on hover or click.

Internal open questions are shown in italics.

### Time Series Data
**Location Dashboard**
The value of a pollutant over time during the specified window.  While locations have varying time intervals over which they report, all time series charts show data at the same intervals.  For one day or one month of data the hourly average is shown.  For the project lifetime the daily averages are shown for the most recent week of data.

**Dataset Dashboard**
The average value of a pollutant over time during the specified window at each individual node selected and the average values across all locations selected.  While locations have varying time intervals over which they report, all time series charts show data at the same intervals.  For one day or one month of data the hourly average is shown.  For the project lifetime the daily averages are shown.  If all locations are selected only the average across all locations is shown, not the individual location values.

### Measurands (Pollutants)
**Location Dashboard**
The average of all values and total number of measurements for each pollutant during the chosen time window.

**Dataset Dashboard**
The average of all values and total number of measurements for the available pollutants during the chosen time window and for the selected locations.  Keep in mind that not all locations may report the same pollutants.
_What are we doing when the locations aren't reporting the same pollutants?_

### Temporal Coverage
 **Location Dashboard**
The average number of measurements for each pollutant by hour, day, or month.  In some views a window may be turned off if that view is not applicable to the selected time window.

**Dataset Dashboard**
The average number of measurements for each pollutant by hour, day, or month at the selected locations.  In some views a window may be turned off if that view is not applicable to the selected time window.

jvntf:
closes #495


necoline (approved):
Looks good! I don't have access to the wireframe, so I'm assuming the styles align - I think they look good. 

I changed the base branch to `testing` since the client is looking at staging and I think we want to have a landing place for changes before we deploy.

PR #520, opened 2021-01-06
Add/518 readme

### LINKED ISSUE #518: Add Technical README to Details Component of the Dashboard
## Goal
It is really important to EDF for the link to the technical README to be easily accessible, so we would like to add it at the bottom of the Details component.  

## Details
This is not shown on wireframes. I'd display the words "Technical README" and have them link to the endpoint that is listed in the API.
### Example
`https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/locations/11639` -> `results[sources[readme]]`

jvntf:
closes #518 

At time of comment, project objects are missing some properties, once added readme's will be rendered there as well.

Location can be tested with
http://localhost:3000/#/location/11639


app/assets/scripts/components/dashboard/details-card.js
```diff
@@ -20,6 +22,19 @@ export default function DetailsCard({ measurements, date, coords }) {
...
+            {sources &&
+              sources
+                .filter(s => s.readme)
+                .map(s => (
+                  <SmartLink
+                    key={s.id}
+                    to={`${config.api}/${s.readme}`}
+                    title="Suggest a new source"
+                  >
+                    {`${s.name} Technical Readme`}
```
AliceR, line 35: The displayed `source.name` feels like a duplication from the source displayed in the sources-card. Maybe label it with Technical Readme only? 
```suggestion
                    Technical Readme
```

app/assets/scripts/components/dashboard/details-card.js
```diff
@@ -20,6 +22,19 @@ export default function DetailsCard({ measurements, date, coords }) {
...
             <CardSubtitle className="card__subtitle">Measurements</CardSubtitle>
+
+            {sources &&
+              sources
+                .filter(s => s.readme)
+                .map(s => (
+                  <SmartLink
+                    key={s.id}
+                    to={`${config.api}/${s.readme}`}
+                    title="Suggest a new source"
```
AliceR, line 33: This title is misleading, maybe
```suggestion
                    title="Get technical details"
```
?

app/assets/styles/global/_content.scss
```diff
@@ -110,3 +110,14 @@
   text-transform: uppercase;
   color: $base-color;
 }
+
+/* Lists
+   ========================================================================== */
```
AliceR, line 115:
```suggestion
/* Info button icon
   ========================================================================== */
```
?

app/assets/scripts/components/dashboard/details-card.js
jvntf: @AliceR I agree it is verbose but i think there is a possibility of having multiple sources and therefore multiple readme's (although as of now perhaps that does not exist). what do you think?

AliceR: I don't know enough about the concept of technical readme's to answer that. Maybe @rwegener2 can help?

rwegener2: I agree with @jvntf that for now we should assume that we could have multiple sources and readmes.  We might change that in response to EDF's feedback next week but for now I think it the path that makes the most sense.

rwegener2:
A few comments:
* In the links I checked there was an extra `v2` in the link so they didn't render.  Ex. `https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2//v2/sources/readme/bay_area_mobile_analysis` was what I was linked to, but it worked if I just adjusted the link to `https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/sources/readme/bay_area_mobile_analysis`
* Right now it looks like this is implemented only on the Datasets page.  Could we also do this for the Locations?

jvntf:
@rwegener2 the readmes are there on locations for sources that have them. see http://localhost:3000/#/location/11639

readme issue is fixed.

rwegener2 (approved):

PR #521, opened 2021-01-06
Update/515 param name

### LINKED ISSUE #515: Display PM25 with the decimal point, as PM2.5, across the site (Parameter filter for Dashboard and Location and Dashboard pages Latest Measurements, Time Series/Temporal Coverage Tabs & Parameters table)

jvntf:
closes #515


app/assets/scripts/components/map/legend.js
```diff
@@ -15,48 +15,48 @@ export default function Legend({ parameters, activeParameter }) {
...
+              <span>{param.displayName}</span>
             </a>
           </li>
         ))}
       </ul>
     </Dropdown>
   );
 
-  const scaleStops = generateLegendStops(activeParameter.toLowerCase());
+  const scaleStops = generateLegendStops(activeParameter.name.toLowerCase());
```
AliceR, line 51: `Uncaught TypeError: Cannot read property 'toLowerCase' of undefined`

app/assets/scripts/views/locations-hub/filter.js
```diff
@@ -48,7 +48,7 @@ export default function Filter({ countries, parameters, sources }) {
   sortList(countries);
   sortList(parameters);
 
-  const parameterNames = [...new Set(parameters.map(p => p.name))];
+  //const parameterNames = [...new Set(parameters.map(p => p.name))];
```
AliceR, line 51: the [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) ensures we have unique parameters in the filter dropdown

app/assets/scripts/views/locations-hub/location-card.js
```diff
@@ -21,6 +21,7 @@ export default function LocationCard({
 }) {
   let updated = moment(lastUpdated).fromNow();
   let started = moment(firstUpdated).format('YYYY/MM/DD');
+  console.log(parametersList);
```
AliceR, line 24:
```suggestion
```

app/assets/scripts/views/locations-hub/filter.js
jvntf: @AliceR I realized that the parameter list contains params with duplicate name but different id and other info (e.x. CO mass / CO Conc both have name === co). I will add a filter by id

jvntf: i think actually that we do not have duplicates (it seemed that way earlier because we were only using name which is non-unique), @bitner can you please confirm?

bitner: parameter could still have duplicates, right now parameterId is the only thing guaranteed to be unique (or the combination of parameter and unit) -- that being said, displayName I believe is now unique and I could enforce that

app/assets/scripts/views/location/nearby-locations.js
```diff
@@ -15,6 +15,7 @@ export default function NearbyLocations({
   parameters,
   activeParameter,
 }) {
+  console.log(activeParameter);
```
necoline, line 18:
```suggestion
```

app/assets/scripts/components/map/legend.js
```diff
@@ -15,48 +15,48 @@ export default function Legend({ parameters, activeParameter }) {
     history.push(`${location.pathname}?parameter=${parameter}`);
   }
 
-  const parameterNames = [...new Set(parameters.map(p => p.name))];
+  //const parameterNames = [...new Set(parameters.map(p => p.name))];
```
necoline, line 18:
```suggestion
```

app/assets/scripts/views/projects-hub/filter.js
```diff
@@ -35,7 +35,7 @@ export default function Filter({ parameters, countries, sources }) {
   );
 
   parameters.sort((a, b) => a.name.localeCompare(b.name));
-  const parameterNames = [...new Set(parameters.map(p => p.name))];
+  //const parameterNames = [...new Set(parameters.map(p => p.name))];
```
necoline, line 38:
```suggestion
```

app/assets/scripts/components/map/legend.js
necoline: I'm still getting this error

necoline (changes requested):
The url params on 'locations-hub' and 'projects-hub' do not update properly. When removing a filter using the pill buttons, it adds the id to the param rather than removing it

I get an error on the country page. I t

necoline (changes requested):
the url params on projects and locations hub are broken when removing them. The id is added to the param a second time rather than removing it

The project cards in projects-hub need updated params - it is still using PM25 for example. The filtering should also be checked again, once this is resolved.

app/assets/scripts/components/map/legend.js
jvntf: this was on country page right? @necoline

AliceR (approved):
With eda77dd I added a small change to this PR that fixes the url problems for me. 
I'll merge the PR into testing, as I need those changes for other work. We can always revisit issues with separate PR's, let's just make sure to note them down on the project board.

PR #523, opened 2021-01-07
Compare locations

danielfdsilva:
In this PR:
- Had some problems with sass when setting up the project, so had to update some modules. I think I managed to solve the problem but I'd like to know if works for everyone else.
- Add back the compare page using the new `/averages` endpoint to display data on the chart.
- Fix the header cards on the homepage using the new `?order_by=random` option and display average measurements like the compare page.
- Convert the `Home` component to functional with hooks.
- Add a "No listed city" default when there's no city available - on the compare page, hub cards and single location.


rwegener2 (changes requested):
* home page loads much faster (and doesn't crash) 🎉 
* thanks for suggesting doing something about the lack of city name Daniel.  I think having "no listed city" on the cards is clearer and cleaner.
* Where it says _"Comparing < dropdown > values for local times"_ would it be easy enough to limit that list to only the pollutants available at one or more of the stations being used in the comparison?  There are a lot more variables now so listing them all means the majority of the time when you select a parameter it isn't available.
* On a similar note to the point above, could we add a list of the parameters available below the city name?  Something like "Reporting: PM2.5 mass, CO2, ..."
* All but one of the stations I tried worked. The one that gave me trouble was:
`http://localhost:3000/#/compare/Alicante%20Lane/US%20Diplomatic%20Post%3A%20Dhaka?parameter=28`.  I think it was that second on in Dhaka that was causing the issue.  Here's my developer tools error:
Screen Shot 2021-01-07 at 5 08 01 PM

In all thanks for bringing this component into the interface and upgrading to hooks! 🚀

app/assets/scripts/actions/compare.js
```diff
@@ -72,6 +78,41 @@ export function fetchCompareLocationIfNeeded(index, location, filters = {}) {
...
+export function fetchRandomCompareLocs(limit = 2) {
+  return dispatch => {
+    // Dispatch all requests.
+    for (let index = 0; index < limit; index++) {
+      dispatch(requestCompareLocation(index));
+    }
+
+    const f = buildAPIQS({
+      order_by: 'random',
+      parameter: 'pm25',
```
necoline, line 91: These names are currently being updated. You can pass the parameter id here instead and that will remain constant

necoline (commented):
It's great to see so many updates to hooks 🎉 

It's a pretty large PR, so I mostly skimmed through it and checked for functionality.

The sass upgrades work for me.

danielfdsilva:
@necoline @rwegener2 Feedback addressed:
- Values in Compare dropdown must be present in at least one locations
- Added "Reporting..." line to the compare location header
- Fix error handling
- Use parameter id instead of name
- Fix conflicts with "testing" branch

AliceR:
@danielfdsilva What do you think about splitting those huge files into smaller components? You can see how we did it with the other views. Right now they are so big Github even hides the diff! This makes reviewing as well as future maintenance more difficult than it should be.

danielfdsilva:
@AliceR agreed. done

AliceR (approved):
briefly tested the branch running locally and everything looks good. 

@rwegener2 not sure about the url you posted above, but I think you should use the location numeric id instead of the name in the url, like http://localhost:3000/#/compare/35822?parameter=130 - was this an old url you stored somewhere in your browser history, or did you access it by following a link in the ui?

danielfdsilva:
@AliceR That's the old url style. Now the locations are referenced by ids

PR #525, opened 2021-01-07
Fix/524 single day

### LINKED ISSUE #524: [Bug] Dashboards - time series and temporal coverage components error with single day selection

jvntf:
closes #524

### DIFF
--- a/app/assets/scripts/components/dashboard/temporal-coverage-card.js
+++ b/app/assets/scripts/components/dashboard/temporal-coverage-card.js
@@ -100,7 +100,7 @@ export default function TemporalCoverageCard({
         ...state,
         [temporal]: { ...state[temporal], fetching: true, error: null },
       }));
-      const [year, month, day] = dateRange ? dateRange.split('/') : [];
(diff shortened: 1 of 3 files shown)


necoline (approved):

rwegener2:
The graph is definitely working better.   Thanks also for pointing about that the temporal coverage charts don't make sense.  I totally missed that detail.
A few comments:
* For me when I first click a date the graph doesn't render for me.  I just get a single dot.  The graph does work fine but I have to refresh the page first.  Can we either have the data update when someone selects a date or put a button to refresh the page?
* Could we switch the tic marks along the bottom line of the x axis to label either every hour or every other hour?  Right now I'm just seeing the Month + day as single tic mark on the far left side of the graph (Ex. Feb 16 -> Feb 17 but nothing in between with a time)
Screen Shot 2021-01-07 at 8 49 02 PM

AliceR:
I pushed 6022557 which should fix @rwegener2's problems with updating the project dashboard cards.

AliceR (approved):
Axis labels might get addressed with #526, let's revisit once both changes are merged.

PR #526, opened 2021-01-07
Timeseries chart

danielfdsilva:
Review after https://github.com/openaq/openaq.org/pull/523. Had to merge from `feature/compare` otherwise the homepage wouldn't load.

Addresses the label issues on the Timeseries chart. Example:
image

image


rwegener2 (changes requested):
* (y-axis of time series) I like that you added the name of the pollutant and the unit.  On my screen the unit gets cuts off.  Can we either add the unit in a new line or increase the height of the time series table?
Screen Shot 2021-01-07 at 4 34 21 PM

* (All labels) Can we make the text bigger?  They sort of blend in to me.
* For the y-axis label on Temporal Coverage, can we write "Count" instead of "Total"?  Because it is technically an average of counts I think the word total might be misleading.

Thank you!

danielfdsilva:
@rwegener2 Addresses feedback:
- It seems it is not possible to add multiple lines on axis labels, so i added the unit to the first vertical tick
- Font size increased
- Replaced "Total" with "Count"

rwegener2 (approved):
All updates look good.  The one issue I had was that the unit still shows up funny for me (Pushed into the number labels).  
Screen Shot 2021-01-08 at 7 40 01 AM
It seems like this may take a little more investigating to figure out something that will look nice so let's merge as is for now (EDF will be happy we put on some kind of label) and come back to it next week.

app/assets/scripts/components/line-chart.js
```diff
@@ -20,31 +20,53 @@ export default function LineChart({ data, xUnit }) {
...
       ],
 
       xAxes: [
         {
           type: 'time',
           time: {
+            tooltipFormat: 'MMM D, YYYY',
             unit: xUnit || 'day',
+            displayFormats: {
+              day: 'MMM D, YYYY',
```
jvntf, line 55: Perhaps this should be formatted as MM/DD/YY to save space? current formatting seems crowded and repetetive.

jvntf (changes requested):
+1 for the unit overlap, other than that looks great

Issue #472, opened 2020-12-14
Map rendering of data locations


rwegener2:
## The Task
Currently no locations are rendering on [the world map](https://openaq-staging.org/#/map).

This map originally showed all of the OpenAQ sensor locations.  It should show those and also show all of the new locations that have been added as a part of the LCS project.

## Notes/Concerns
@bitner Do you foresee any backend concerns?

### Mobile Nodes
I believe that mobile nodes come back as a list of points.  Is there a way to render those as a single point for simplicity?  It seems like at most zoom levels this would be preferred.

## Updated Styling
We would like at some point for LCS sensors to render as squares and government sensors to render as circles, something like in this wireframe:
![Screen Shot 2020-12-13 at 8 33 02 PM](https://user-images.githubusercontent.com/35503632/102033956-6ab19d80-3d82-11eb-8c82-6bb6b7fb0fb2.png)

Devs should decide, however, if that is easier to do as a part of this ticket or a separate one.  The styling updates are currently documented in [this card](https://github.com/openaq/openaq.org/projects/1#card-50060787).

AliceR:
https://github.com/openaq/openaq.org/pull/476 contributes to this ticket by updating the maps to use vector tiles which allows us to display _a lot more_ locations than with the previous approach of fetching and parsing locations from endpoint to display as geojson. We also added a layer for the mobile sensor bboxes.
We still would need to:
- scale the old visualization approach for more parameters (currently only a small set of parameters is handled in [utils/map-settings.js](https://github.com/openaq/openaq.org/blob/develop/app/assets/scripts/utils/map-settings.js)), all the new ones need to be included in the approach)
- differentiate the symbolization for lcs vs government
- filter the locations on the country page to only show those, as in the old version :
<img width="1502" alt="Screenshot 2020-12-21 at 20 00 53" src="https://user-images.githubusercontent.com/1732578/102812411-3f086600-43c7-11eb-8b79-29f75053a3f3.png">

rwegener2:
I added the three tasks from @AliceR's comment as cards in the project board with references to this ticket.

rwegener2:
One question to investigate on the above comment - After how long without no updates before a location on the existing world map turns grey?  OpenAQ didn't know off hand what this length of time is, but they said they wanted it to be roughly the same as the current world map functionality.  They thought this would be roughly a few days.  For the "confirm date functionality" task above we should determine how long this length of time is.

Issue #513, opened 2021-01-05
Dashboard pages visual updates


rwegener2:
## Goal
Update the dashboard pages in accordance with the lovely designs that Tânia has made ([wireframe](https://projects.invisionapp.com/d/#/console/20815156/437956789/preview)).  Changes should affect both Locations dashboards ([example](https://openaq-staging.org/#/location/31233)) and Datasets (project) dashboards ([example](https://openaq-staging.org/#/project/27597)).

⭐ tasks are the priority tasks for the Friday milestone

## Sub-tasks
**Top components**
- [x] Change top three components (Details, Latest Measurements, Sources) to grey
- [ ] ~Details component - change the font sizes~
- [x] ⭐ Details component (AQDC datasets only) - display Data Stage (Ex. `https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/locations?sourceName=bay_area_mobile_analysis` @ `sources` -> _index_ -> `lifecycle_stage`)
- [ ] ~Details component - add lcs or gov and mobile or stationary tags in orange~

**Time series**
- [x] ⭐ add unit to the vertical axis
- [x] ⭐ add the label & year to the horizontal axis (this isn't shown in the wireframes - I'd just write "Date" as the label and put the year in parenthesis)
- [x] add unit to the parameter value in the hover text
- [x] remove the lines connecting the individual dots except when hovering over a dot

**Other**
- [x] responsiveness - Latest Measurements number, time series parameter tabs, and temporal coverage parameter tabs overlap when size of the screen is changed ([example](https://openaq-staging.org/#/location/38070))
- [x] Parameters table described below in the [comment](https://github.com/openaq/openaq.org/issues/513#issuecomment-756246567)

### Dataset only sub-tasks ([dataset wireframe](https://projects.invisionapp.com/d/#/console/20815156/437956796/preview))
- [x] Details card - add # of Locations above # of Measurements

### Optional (not project critical)
- [ ] Time Series - pretty up (add color to) the hover value
- [ ] Details - add the little icons to the dates and lat/lon

rwegener2:
@danielfdsilva If it fits in it would also be great if you could add units to the averages in the Parameters table as well.
So something like:
Parameter | Avg | Count
| :------------- | :----------: | -----------: |
PM2.5 | 300 ug/m^3 | 700k
CO conc | 8.23 ppm | 503k
This can supersede The AQDC data stage if you are short on time.  Thanks!

[...]

PR #527, opened 2021-01-07
More map features

AliceR:
Contributing to #472 :
- Limit parameters on world map to hardcoded list in map-settings: not a fan of this approach, but in the same time not sure what a better alternative could be? Since there is the idea of "original" parameters vs. those measured by recently added sources, we need a way to specify which parameters we want to be able to select from the dropdown in the legend. Bonus: one more component without redux.

- Filter map locations by country on country page: using the property `country` provided with the vector tiles

### DIFF
--- a/app/assets/scripts/components/map/legend.js
+++ b/app/assets/scripts/components/map/legend.js
@@ -79,5 +79,9 @@ export default function Legend({ parameters, activeParameter }) {
 
 Legend.propTypes = {
   parameters: PropTypes.array.isRequired,
-  activeParameter: PropTypes.object,
+  activeParameter: PropTypes.shape({
+    displayName: PropTypes.string.isRequired,
+    name: PropTypes.string.isRequired,
+    id: PropTypes.number.isRequired,
+  }).isRequired,
 };
--- a/app/assets/scripts/components/map/legend.js
+++ b/app/assets/scripts/components/map/legend.js
@@ -46,7 +46,7 @@ export default function Legend({ parameters, activeParameter }) {
     </Dropdown>
   );
 
-  const scaleStops = generateLegendStops(activeParameter.name.toLowerCase());
+  const scaleStops = generateLegendStops(activeParameter.id);
   const colorWidth = 100 / scaleStops.length;
 
   return (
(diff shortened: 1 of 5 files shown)


jvntf (commented):
Looks good! I don't know all the details of the 'original' parameters conversation, but perhaps we can just include a flag in the list returned by parameters endpoint ('tileable' or something like that). ParameterConversion factor could also be included.

jvntf:
also latest features have been getting merged to `testing`, perhaps this PR should use that as base?

jvntf:
also looks like this is branched from develop, but just a heads up that latest features have been getting merged to `testing` so might be good to merge that and use it as base.

rwegener2:
This looks nice.  The color scales are operating much more smoothly.  I noticed that some of the units are getting conflated.  For example with CO the stations in the US measuring ppm show up blue while the stations in Europe measuring in ug/m^3 show read because their values are in the hundreds.  On (I believe) the staging API there is a new setup for parameters where the ppm measurements should belong to a parameter `CO conc` and the ug/m^3 belong to a parameter `CO mass`.  The values that should be shown on the OpenAQ map will be the ones labeled "conc" (Ex. BC conc, CO conc, NO2 conc, etc.).  As a general rule we are also denoting parameters by their `displayName`.  I would think that if you merged to testing and used the staging API that you could iron out anything related to that change in parameters (which hopefully isn't too much 🤞 )

One other note about on Countries page:
Something seems a little off with the "locations" number listed on the countries cards.  For some countries, such as Costa Rica (http://localhost:3000/#/countries/CR) locations is 1 but I see no stations.  For others (Germany http://localhost:3000/#/countries/DE or Canada http://localhost:3000/#/countries/CA) the number locations listed is far fewer than the number of points on the map.  What is the source of the "locations" number being displayed on the countries cards right now?
We can fix this in a different PR but I just thought I'd ask the question here since I'm noticing it.

AliceR:
Thanks @jvntf & @rwegener2 , good to hear the parameters are sorted out and provided api side. Sorry that I missed that! I'll rebase onto `testing` and clean up my code 👍 

About the location count, @rwegener2 could you try to scroll down on the country page and see if more locations come up? On small screens it looks like the first section fills the page completely, but there might be more sections. Might be worth a "visual update" along with #513  ?

rwegener2:
@AliceR I added [a card](https://github.com/openaq/openaq.org/projects/1#card-52436897) for the # of locations investigation.  So that issue is now separate from this PR.  It doesn't need to be prioritized.

jvntf (approved):

Issue #483, opened 2020-12-16
Add node selection to the datasets dashboard


rwegener2:
## Goal
Allow the user to select a custom set of nodes for data aggregations, but limit them to 15 nodes or less to keep within the limits of our backend

## Details
The user flow is shown in [this set of wireframes](https://projects.invisionapp.com/share/4PZP8PX96ZU#/screens/440256752?browse). The date selection feature (#481) is also included in that link, but this ticket just refers to node selection.

### User Flow
1. When a user hits a page the default display has "All locations selected".  All locations on the map are shown as selected ([screen](8PX96ZU#/screens/440256752))
2. When in "select locations" is chosen the user can pick individual locations.  A hover lets users select the location. ([screen](https://projects.invisionapp.com/share/4PZP8PX96ZU#/screens/440490712))
3. Once a location is picked all other locations are turned off.  A chip is added to the top and their location tally (up to 15) is updated.  ([screen](https://projects.invisionapp.com/share/4PZP8PX96ZU#/screens/440256753))
4. A user can select up to 15 locations.  After that no more locations can be added.  ([screen](https://projects.invisionapp.com/share/4PZP8PX96ZU#/screens/440490713))
5. The data updates when the user hits the "Apply selection" button

### Dashboard Data Display
Once the "apply selection" button is hit the following components should to aggregate over only the selected locations:
* time series - the time series would do two things with node selection 1) it would show the average of the nodes 2) it would show each node's points individually.  So on a page with 5 nodes selected there would be six lines on the graph - one per node and then one for the average.   The exception to this would be when all nodes are selected.  In this case only the average should be shown.
* measurands
* temporal coverage

See #481 for more details about how the data should already be changing in accordance with the date.

PR #528, opened 2021-01-08
Feature/node selection

necoline:
Closes https://github.com/openaq/openaq.org/issues/483

- adds map to projects page that displays locations and centers on project country
- updates bbox for US 
- adds selection of locations
- triggers fetch of location data and uses those params to display in the dashboard

Todo:
- [ ] highlight selected location in some way
- [x] fix the the request to access project data that is specific to a set of locations -- needs attention on the backend, but the frontend is set up
- [ ] add tests
- [ ] close popup once 'add location' is selected - right now, it stays up and creates confusion with the pill does not update immediately
- [x] adds legend dropdown to allow for viewing of all locations
- [x] fixes location payer to work with array of location ids


AliceR:
This is looking good already!

The location layer component currently is build to only appear once - it has static layer ids (`id: 'location-layer'`). If you want to reuse it for multiple locations, you would need to either make the layer id dynamic based on the location id, or - which I think would be the better option - make the filter accept an array of location ids.

PR #532, opened 2021-01-08
Improve table component and add parameter unit

danielfdsilva:
While looking at adding the parameter unit (#513 ) to the table I came across some problems which were complicating the task. Given that the table was not being used anywhere else I refactored it.

- The table had a mix-and-match of styles, some coming from styled components, other being added as properties. The openaq-design-system has styles for tables which are more in sync with the rest of the site. I used these instead.
- The sort was not working properly. It was being done on the rendered value which is a string. This leads to the wrong order when ordering numbers.
- Adding the unit as part of the value would just compound the problem above. The new behavior is to sort by a raw value (the number) and then render a display value.

### DIFF
--- a/app/assets/scripts/components/dashboard/measurands-card.js
+++ b/app/assets/scripts/components/dashboard/measurands-card.js
@@ -3,35 +3,31 @@ import { PropTypes as T } from 'prop-types';
 
 import Card from '../card';
 import Table from '../table';
-import { shortenLargeNumber, round } from '../../utils/format';
+import { shortenLargeNumber } from '../../utils/format';
 
-const initData = {
-  parameter: {
-    values: [],
-    formatHeader: v => v.toUpperCase(),
-    style: {
-      color: 'black',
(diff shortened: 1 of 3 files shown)


rwegener2 (approved):
Thanks @danielfdsilva for all your improvements to the codebase with this task!  I like the way the new table looks, too.

It looks like a few of the values - `PM0.3 count` and `PM0.5 count` are showing up with the "k" in the number for thousand in some cases (ex. http://localhost:3000/#/location/31787) because their values are in the thousands.  I don't love this but I think since nothing is broken I will add this as a small fix to our board so we can close this PR out as is.

jvntf (approved):
I believe the test is failing due to cypress timeout while waiting for data to load.

danielfdsilva:
@rwegener2 that's the behavior of the `shortenLargeNumber` function. If this is not desirable across the website we can easily change this. Let me know

PR #536, opened 2021-01-08
Fix filter display

danielfdsilva:
Correct the display of filter dropdowns to match others.


jvntf (changes requested):
Looks good, will you also add a style override to make text and outlines black per the wireframes? (unless this has changed, im not sure where the most up to date designs are)
image

danielfdsilva:
@jvntf Done. Now using the base button style. It is not exactly the same as the image you share, but it's the style coming from the design system

jvntf (approved):
Added a commit to patch the tests

PR #538, opened 2021-01-08
Update/531 nomenclature

### LINKED ISSUE #531: Nomenclature change in hub cards and dashboard headers
## Goal
There have been lots of discussions about nomenclature but decisions have been made.  The two updates to make are:
* change the word "Government" to "Reference-Grade".  
* mentions of "Lcs" should display the full phrase "Low Cost Sensor"

## Details
These changes will be relevant in the orange tags on the cards of the [Locations Hub](https://openaq-staging.org/#/locations)
Screen Shot 2021-01-07 at 9 01 49 PM

and in the headers of the locations dashboards ([example](https://openaq-staging.org/#/location/145))
Screen Shot 2021-01-07 at 9 03 31 PM

jvntf:
closes #531 

The api encodes source type as 'lcs' and 'government'. If theres no reason why these can't change (not sure if this is the case), i think its a better solution to just change the values instead of mapping them on the front end.


AliceR:
I agree, let's chat with @bitner to update the api accordingly

jvntf:
per this mornings convo, this will be updated using the properties in sources objects @AliceR. reverting to draft

jvntf:
@AliceR @necoline this is updated

rwegener2:
This PR looks nice to me.  Can we just change the word "Reference" to the full "Reference grade"?

If it is possible it would nice to get this PR merged sooner rather than later.  I think it will help us switch our vocabulary and make it less confusing when the new set of sensor type filters get added.

jvntf:
@rwegener2 'Reference' needs to be changed to 'Reference-grade' on the backend cc @bitner

rwegener2:
Okay.  If the frontend will automatically change with the backend change I think this can be merged (@jvntf once you are happy with code reviews)

Issue #534, opened 2021-01-08
Location dashboard page - Sources update


tania-pires:
## Goal
Clarify how technical readmes are associated with sources as well as specify the lack of a readme by changing how the readme is displayed.

### Background
* A technical readme is associated with a specific source, but not all sources have one.  
* In the majority of cases there is only one source per location or dataset.  However, we want to be able to handle the case of having more than one source.
* There are sometimes (but not always or even often) not source links.  The current link on the Source name seems to link back to the OpenAQ homepage.

### Implementation
API examples using `https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/locations/11644`
- [ ] move the technical readme link from the Details box to the Sources box and place the link below the respective source name
Technical readme @ `results` -> `sources` -> _index_ -> `readme`
- [ ] include the text "No technical readme for this source" in place of the link if there is no readme
- [ ] update the sources to link to their specified url when available.  Don't link anywhere when there is no source link
Source link @ `results` -> `sources` ->  _index_ -> `url`
- [ ] (Nice to have) use the icon to demonstrate to the user when there is a source link.  If the icon is time consuming let's still use something like an underline to show clickability

## Screenshot
<img width="443" alt="Screenshot 2021-01-08 at 15 05 58" src="https://user-images.githubusercontent.com/22524740/104030259-050ad280-51c3-11eb-95e3-3a2af68ce348.png">

jvntf:
@danielfdsilva i assigned you because seems like you are working on this already

PR #540, opened 2021-01-11
Update location sources display

danielfdsilva:
Apply the updates listed in #534

### DIFF
--- a/app/assets/scripts/components/dashboard/sources-card.js
+++ b/app/assets/scripts/components/dashboard/sources-card.js
@@ -9,8 +9,10 @@ import config from '../../config';
 const SourceList = styled.ul`
   padding: 0;
   list-style: none;
-  display: grid;
-  grid-gap: 0.5rem;
+
+  > *:not(:last-child) {
+    margin-bottom: 0.5rem;
+  }
 `;
 
 const InfoMessage = styled.div`
@@ -23,7 +25,7 @@ export default function SourceInfo({ sources }) {
   return (
     <Card
       title={sources && sources.length > 1 ? 'Sources' : 'Source'}
-      gridColumn={'11 / -1'}
+      gridColumn={'10 / -1'}
       renderBody={() => {
         if (sources) {
           return (
@@ -36,11 +38,14 @@ export default function SourceInfo({ sources }) {
                       href={source.url}
                       rel="noreferrer noopener"
                       target="_blank"
+                      title={source.name}
                     >
                       {source.name}
                     </a>
                   ) : (
-                    <p className="source__title">{source.name}</p>
+                    <p className="source__title" title={source.name}>
+                      {source.name}
+                    </p>
                   )}
                   <p className="source__readme">
                     {source.readme ? (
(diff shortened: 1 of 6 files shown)


rwegener2 (changes requested):
Visually this looks really nice.

When a user clicks a link, either the Source link or the Tech readme, can it open in a new tab?

If it's helpful to other folks reviewing here is a link to a location that has both a technical read me and a source link http://localhost:3000/#/location/11644

danielfdsilva:
@rwegener2 addressed

app/assets/styles/locations/_single.scss
```diff
@@ -148,8 +138,33 @@
   }
 }
 
-/* Fold: Dataset
-   ========================================================================== */
+.source__title {
+  color: $primary-color;
```
AliceR, line 142: Not sure, but I'd think this should only be blue if clickable? Should the coloring move to `--external`?

app/assets/scripts/components/dashboard/sources-card.js
```diff
@@ -29,13 +29,36 @@ export default function SourceInfo({ sources }) {
...
+                          /^\/v2\//,
+                          ''
+                        )}`}
+                        rel="noreferrer noopener"
+                        target="_blank"
+                      >
+                        Technical readme
+                      </a>
+                    ) : (
+                      'No technical readme for this source'
```
AliceR, line 58: If there is no technical readme, should we just skip it instead?

AliceR (approved):
Looks good, only few thoughts added below

app/assets/styles/locations/_single.scss
AliceR: Thanks to @rwegener2's location shared above I understood that the source name is never clickable, there will be an extra field displayed. (Still, I find it confusing that a non-clickable element has the same color as a link usually has)

danielfdsilva: According to the screenshot on the issue, the differentiation lies in the icon. @tania-pires is this correct?

app/assets/scripts/components/dashboard/sources-card.js
danielfdsilva: This was specifically requested in the issue.

rwegener2: Yea, I guess I could go either way.  I added that to the issue because I kept forgetting that only a small number of datasets have technical readmes.  I kept getting frustrated looking for them and wanted a lack of readme to be specified.  Users may not be actively looking for them, though, so perhaps the text is unnecessary.  Keep the "no readme" text or not @danielfdsilva you can make the final call.

app/assets/styles/locations/_single.scss
rwegener2: I agree the coloring is a little confusing.  Seems like this would be more than a minor fix, though?  We would either be changing the color of all the light blue text on the website (or at least the dashboard page) or all the links.

rwegener2 (approved):
I'm approving because I think the functionality looks good.

I am curious to hear what @tania-pires has to say about the link colors.  If there were something simple to do I'd be all about it.

danielfdsilva:
After consulting with @tania-pires, I made a couple of changes to the sources and structure:

- Title is now black when there's no link
- Very long titles are truncated and the full title is displayed on hover
- The sources card size was increased a bit
- The measurements list was fixed to prevent overlapping and add some space

image

@rwegener2 requesting a new review

rwegener2 (approved):
This looks great.  I love that you've also fixed the overlapping Latest Measurements values.

Issue #537, opened 2021-01-08
Announcement banner on homepage


rwegener2:
## Goal
OpenAQ would like to announce their exciting new data on their homepage.  We have wireframes for what that could look like [here](https://projects.invisionapp.com/share/EZZGACI7CPD#/screens/441697018_1-_Homepage_Banner).

## Details
Of the two links the "Explore the data" one should link to the locations page and the "read our blog post" one should just be empty for now.  We can add the link when OpenAQ has it written.

rwegener2:
Feedback:
> Homepage Banner design from Tania looks great! Is it possible to move it below the white navigation menu and make the bar and text a bit bigger?

tania-pires:
To bar redesign according to feedback: 
![OpenAq-Top bar notification](https://user-images.githubusercontent.com/22524740/104188515-42fd3600-5411-11eb-9989-d33be76f244b.png)

PR #541, opened 2021-01-11
Add news banner

danielfdsilva:
Implemented as outlined in #537

### DIFF
--- a/app/assets/scripts/components/news-banner.js
+++ b/app/assets/scripts/components/news-banner.js
@@ -1,0 +1,37 @@
+import React, { useState } from 'react';
+
+import SmartLink from './smart-link';
+
+export default function NewsBanner() {
+  const [isVisible, setVisibility] = useState(true);
+
+  return isVisible ? (
+    <div className="news-banner">
+      <div className="inner">
+        <span className="news-banner__pill">New</span>
+        <p className="news-banner__msg">
+          We have low cost sensor data! 🙌{' '}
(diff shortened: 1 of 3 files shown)


rwegener2 (approved):
From a functionality perspective this looks fantastic.

We don't have a blog link from OpenAQ yet.  Should we just merge this after code review and add that link later this week when they will have that published?

rwegener2:
Sorry for the scatterbrained comments, feedback is all over the place.

I found this additional request about the banner:
* center the text
* "call out the links in separate and clear buttons"

That second bullet leaves a bit up to interpretation.  Maybe also underlining them or something to show that they are hyperlinks?  I think the hope might be that we also add the oval box around that text, but I think then we probably want to change the color or styling of the "New" text so that then doesn't appear to also be a link.

rwegener2 (changes requested):
Sorry for the scatterbrained comments, feedback is all over the place.

I found this additional request about the banner:
* center the text
* "call out the links in separate and clear buttons"

That second bullet leaves a bit up to interpretation.  Maybe also underlining them or something to show that they are hyperlinks?  I think the hope might be that we also add the oval box around that text, but I think then we probably want to change the color or styling of the "New" text so that then doesn't appear to also be a link.

danielfdsilva:
@rwegener2 Banner restyled to match new design

rwegener2 (approved):
Looks great

PR #544, opened 2021-01-11
Update map symbolization

AliceR:
Contribute to #472 

This is splitting the measurements layer into two symbols, squared and circled.
I am using an `Uint8Array` to create the square and I am not sure why this needs an initial color, it is going to be repainted anyways just like the circles. It would be cool to use something like [this approach](https://stackoverflow.com/a/30585171/1937860) to draw the circles in the same way as squared, just with transparent corners. We could re-use the same layer that way, just referring to another image based on the property.

Currently, I only have a `isMobile` property available on the vector tiles. @bitner , could you add something to differentiate 	      'Low Cost Sensor' vs. 'Reference-Grade'? I need to revise language in the symbol legend as well, can you help @rwegener2 ?

### DIFF
--- a/app/assets/scripts/components/map/measurements-layer.js
+++ b/app/assets/scripts/components/map/measurements-layer.js
@@ -15,6 +15,27 @@ import {
 import { generateColorStops } from '../../utils/colors';
 import Popover from './popover';
 
+const square = {
+  width: 12,
+  height: 12,
+  data: new Uint8Array(12 * 12 * 4),
+  render: function () {
+    const bytesPerPixel = 4;
+    for (let x = 0; x < this.width; x++) {
+      for (let y = 0; y < this.height; y++) {
+        const offset = (y * this.width + x) * bytesPerPixel;
+        this.data[offset + 0] = 0;
+        this.data[offset + 1] = 128;
+        this.data[offset + 2] = 128;
+        this.data[offset + 3] = 255;
+      }
+    }
+
+    // not sure why this needs an initial color, it is going to be repainted anyways
+    return true;
+  },
+};
+
 export default function MeasurementsLayer({
   activeParameter,
   isAllLocations,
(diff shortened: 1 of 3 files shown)


app/assets/scripts/components/map/measurements-layer.js
```diff
@@ -102,43 +185,58 @@ export default function MeasurementsLayer({
...
-    if (country && map.getLayer(`${activeParameter}-layer`)) {
-      map.setFilter(`${activeParameter}-outline`, ['==', 'country', country]);
-      map.setFilter(`${activeParameter}-layer`, ['==', 'country', country]);
-
-      return () => {
-        map.setFilter(`${activeParameter}-outline`, null);
-        map.setFilter(`${activeParameter}-layer`, null);
-      };
+    if (country && map.getLayer(`${activeParameter}-circles`)) {
+      map.setFilter(`${activeParameter}-square-outline`, squaresCountryFilter);
```
jvntf, line 196: image
I think this is not working (at least on location dashboard) because country has not been passed to `MeasurementsLayer`.  We should pass country [here](https://github.com/openaq/openaq.org/blob/feature/472-map-symbols/app/assets/scripts/views/location/nearby-locations.js#L38)

jvntf (approved):
Approving as this looks good. See one small change below that is sort of tangential to this PR.

I think some of the colors here may be confusing. The legend shows an OpenAQ blue circle and square data station, but they appear colored according to their value on the map. Should they perhaps be a very distinct color in the legend? White with black outline? Not a huge deal, just took me a minute to understand that the variable colored shapes correspond to the static colors in the legend.

rwegener2:
This looks nice.  Just one small comment:
* I can't click on the point and have the little card pop up anymore.  Is this just because I'm running locally or was this removed somehow?

As for the symbol legend "Low cost sensor" and "Reference grade sensor" work great.

app/assets/scripts/components/map/measurements-layer.js
AliceR: I intentionally did not filter by country on the location dashboard, I thought we'd want to see all surrounding stations even when across the border. Usually the map would be zoomed in to the selected location and highlight it. Should we limit the zoom range on the location page to not be able to get to the view you shared above? Its a question for @rwegener2 and @tania-pires I'd say

AliceR:
@rwegener2 the popup should still work. Can you specify when it does not? Is it a specific page or location that it happens to not open?

Issue #542, opened 2021-01-11
Filtering bugs - Locations & Datasets Hubs


rwegener2:
## Goal
Ensure that the filters on the Locations and Dashboards Hubs are responding properly.

## Details
Bullets are work items.  Quotes below the items come from the feedback documents.
### Locations Filters
- [x] Make sure data source is working
> I chose United States for the Country and Houston Stationary BEACO2N for the Data Source and get 3500 location results that include PurpleAir. 
- [x] Make sure EDF sources (Ex. source id `burlington_methane`, `houston_beaco2n`, or `nyc_mobile_mit_city_scanner`) are showing up in the search results
> Doesn't seem like low-cost sensor nodes from EDF datasets are showing up on locations search

### Locations and Datasets Filters
- [x] Make sure relevant parameters are available in the dropdown
> Some of the parameters in that filter don’t have associated data (ex. BC conc)
* I also noticed that CO2 and NOX are available as filters, but I don't think we should have those anymore. @bitner Might comment on that.  It could be a holdover from the unit conversions.

### Order By
- [ ] ~I can't tell if any of these are working properly.  I think the reason for that is that null results show up first.  Not sure if this is a frontend fix or something @bitner wants to change on the backend.~
> Count does not seem to be in order of measurement count. For city, location, and country, if any of those are null, those results shows up
- [ ] (Nice to have) I think it is confusing that the order by selection shows up next to the filter selection ([example](https://openaq-staging.org/#/locations?order_by=country&countries=AU)).  If we can display the order by selection on the right hand side or something low lift that would be great. @tania-pires do you have any suggestions here?

## Note
I think we have gone through and fixed this before.  If we were to implement more tests this could be a good candidate item.

jvntf:
Leaving open because final 2 points require backend changes @rwegener2

rwegener2:
I am pretty sure that the first order by item is working now.  I've striked it out in the ticket body to indicate that it seems to be fixed but wasn't intentionally worked on.

PR #545, opened 2021-01-11
Fix location source filter

jvntf:
Address items 1, 2 of #542 cc @rwegener2


AliceR (approved):

rwegener2:
This is definitely functioning better.  When I filter on just data source it seems to work well.

I typed out something last night with some situations where this was working strangely, but now that I've come back to the same ones morning it seems fine.  So I guess we should merge it for now and address those issues if they resurface.

PR #546, opened 2021-01-11
Update sensor type filter

### LINKED ISSUE #535: Locations Hub page - Filters updates
## Goal
The "Sensor Type" filter has gotten more detailed, so we would like to change the format of the filter dropdown so as not to clutter the filter lists.  This will involve adding new filters to the search.

## Details
The "Sensor Type" filter allows 3 distinct selections: Grade, Mobility, and Entity. 

### Grade
Low Cost Sensor or Reference-grade

**Example API Call**
`https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/locations/35840`
`results` -> `sources` -> `sensorType`

**Manufacturer Filter**
The Manufacturer type dropdown will be enabled only if Low Cost Sensor is chosen.

Implementation notes:
I see that if you know the name of a manufacturer you should are able to filter locations by it (Example https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/locations?manufacturerName=Picarro).  I don't see an endpoint to list the manufacturers, however.  @bitner would you please comment how many manufacturers there are across all the datasets?  Could we make an additional endpoint to enumerate the available options?

### Mobility 
Mobile or Stationary

**Example API Call**
`https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/locations/35840`
`results` -> `isMobile`
This can mirror the functionality of the existing mobility filter.

### Entity 
Community or Research or Government

**Example API Call**
`https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/locations/35840`
`results` -> `sources` -> `entity`

## Screenshots
Screenshot 2021-01-08 at 15 20 25

Sensor type

## Notes
The screenshots say "Lcs" but we would like to be more explicit going forward, so let's type out "Low Cost Sensor" instead.

jvntf:
- [x] Setup sensor type filter component
- [x] Add all options
- [x] Connect to filter saving / query string system
- [x] Send to API
- [x] Add proper styling
closes #535


app/assets/scripts/actions/base-data.js
```diff
@@ -80,5 +82,7 @@ export function fetchBaseData() {
     fetcher('countries');
     fetcher('sources');
     fetcher('parameters');
+    fetcher('manufacturers');
```
AliceR, line 85: Why did you decide to add these to the base data?

Personally, I am trying to eliminate the base data / base stats concept, as these requests happen on every page, even the static about page, where we do not at all need this data. Would it be reasonable enough to change the code to only fetch the lists when we need them? Maybe in a useEffect within those components?

app/assets/scripts/views/locations-hub/sensor-type-filter.js
```diff
@@ -0,0 +1,172 @@
+import React, { useState } from 'react';
+import T from 'prop-types';
+import styled from 'Styled-Components';
```
AliceR, line 3: `Javascript error: { Error: Can't walk dependency graph: Cannot find module 'Styled-Components'`

app/assets/scripts/actions/base-data.js
jvntf: Honestly I put it here to maintain consistency with most of the requests that are using the reduxeed api format (fetching, fetched etc). Can add it here as a one of useEffect request but I thought maybe this data was needed elsewhere and I thought it was simpler to add here than to setup a single action/reducer.
As of now the app calls `fetchBaseData` in main.js. We could refactor this call to take params, and introduce it to relevant components. perhaps in separate PR? what do you think?

AliceR: The general idea was to move away from redux, we have useEffect fetches in other places right now. I'll leave it to you to decide if that blows the scope of this implementation.

rwegener2:
This looks great - I like the detail you got in about switching the orange tag if a filter is already selected and also how you set the buttons up to be able to unclick.

A few requests to functionality:
* Can we please only allow the manufacturer filter to be used if LCS filter is selected?
* Can we also add this filter to the datasets page?  If this is a significant lift (since you will have to loop through sources to determine the entity of a dataset) we can do it in a separate PR.  I'll leave it up to you @jvntf if it makes more sense to add that here or to leave this as it is.

jvntf:
@rwegener2 im not sure it this filter can be added to the projects hub. that endpoint doesnt take these query params (as of now).

jvntf:
@rwegener2 first point is now addressed

rwegener2:
1st point looks good.  Second point I moved to a separate card so we can investigate API needs.

To me this is ready to merge.

PR #554, opened 2021-01-12
Improve timeseries chart display

danielfdsilva:
Result:
image


rwegener2 (approved):
Very pretty.  Thanks!

PR #556, opened 2021-01-12
Fix tiles request

AliceR:
Using parameter id instead of name to request tiles. Also changing the default parameter to be pm2.5 (as it was in previous versions)


jvntf (changes requested):
@AliceR im seeing 500 errors on the tile requests and no data in the map

AliceR:
@jvntf this does not seem to be related to this change

jvntf (approved):

PR #559, opened 2021-01-12
Fix api breaking change

jvntf:
Update after api response format change for `sensorType`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-# OpenAQ 
-
+# OpenAQ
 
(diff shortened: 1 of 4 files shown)


rwegener2 (approved):
The changes relevant to sensor type look good to me.  

The time series charts on all the dashboards I clicked on are quite strange - they show multiple readings for each point in time - but I am pretty sure that wouldn't be caused by anything in this PR.  @jvntf if you agree that none of the changes here would likely cause the time series issue I think it looks good to merge.

AliceR (approved):

PR #560, opened 2021-01-12
TEMP patch country page request, needs api change

jvntf:
Fixes request to /countries endpoint


jvntf:
No longer needed, bug getting fixed on backend

PR #561, opened 2021-01-13
Update/node selection

### LINKED ISSUE #483: Add node selection to the datasets dashboard

necoline:
- pulls data from averages endpoint and creates a single aggregate object for each parameter and location selected by the user
- splits the project dashboard into two options - initial project dashboard and dashboard with calculated averages
- adds a new timeseries table that doesn't make a call to the /averages endpoint since that is already done prior to rendering the chart -- this is slightly duplicative, but I thought it was most readable

Known Issues: some projects get 500 errors when selecting locations (ei 27581) I'm leaving an error message here instead of spending time tracking this down

To test:
UI - 
- select a project from the project hub, I like 27595 since it has multiple params to chose from
- in the map check 'select locations'
- click on locations and click 'select' add as many as you want, switch the params using the legend and add some from there
- select 'Apply Selection' and review the dashboard

Code: 
- I recommend reviewing the following to make sure they make sense and are readable
 -  `app/assets/scripts/views/project/nodes-dashboard.js`
 -  `app/assets/scripts/views/project/project.js`
 - `app/assets/scripts/components/dashboard/node-time-series-card.js`

this fully closes #483 with the exception of displaying all the locations as well as the averages on the timeseries card


app/assets/scripts/views/project/nodes-dashboard.js
```diff
@@ -0,0 +1,220 @@
...
+    return (
+      <ErrorMessage
+        isShowingDiagnosis={false}
+        instructions="Please try a different location or parameter"
+      />
+    );
+  }
+
+  return (
+    <div className="inner dashboard-cards">
```
AliceR, line 167: it looks like we should be able to reuse the `project-dashboard` here

app/assets/scripts/components/dashboard/node-time-series-card.js
```diff
@@ -0,0 +1,86 @@
...
+  padding: 0;
+`;
+
+const CardHeader = styled(BaseHeader)`
+  display: grid;
+  grid-template-rows: min-content 1fr;
+  grid-gap: 0.5rem;
+`;
+
+export default function TimeSeriesCard({ parameters, data, titleInfo }) {
```
AliceR, line 26: Why did you duplicate the component? We can reuse the card and chart component for both time series data sources, fetched and generated

AliceR (commented):
I have a couple of small UX issues:
- project http://localhost:3000/#/project/27593 crashes when trying to `select nodes`
- the apply selection button seems to be skipped, the page reloads when I select locations - which I think is fine, we can just drop the button?
- clearing the selection still maintains it for when I want to select new locations - I would need to manually find them on the map to remove via popup

 I would not mind shipping what you have and write tickets for the smaller tasks remaining.

rwegener2:
Nice work @necoline with a complicated feature.

Just a few "nice to have" comments:
* I was a little unsure playing with this if the values updated because you can't see the values from where the "apply selection" button is.  I think it would be nice if the button changed to grey after you hit update and then went back to blue when you changed something, but if that would take more than 20-30 minutes I think this can be skipped.
* It takes me a second to find the "All locations" and "select locations" choice.  Can we make this text a smidge bigger?

One other thing I noticed:
* at least when I was playing with this the names of the selected locations did not appear to be showing up. 
 What might be happening is that the names are showing up but as empty strings so maybe the colored text just doesn't look like anything?  Not sure if this makes sense.  I say that because at the time of writing clicking on a map location leaves the location name blank in the pop up.
Screen Shot 2021-01-14 at 10 25 31 AM
(Image: no indication of location names next to 2/15 and no name of location in the map popup)

I did all of my experimenting on http://localhost:3000/#/project/27585

I'm 💯 with Alice - down to merge and then fix the smaller issues in separate tickets.  That might also help spread out the workload since I think @necoline you have been managing this total beast of a ticket mostly by yourself

app/assets/scripts/components/dashboard/node-time-series-card.js
necoline: There is definitely a cleaner way of handling this, but the timeseries accesses the /averages endpoint with either a project id or a single location. The response for passing in multiple locations as a group is different than the prior cases and needs to different logic for the card and chart to consume it properly. 

Since I am already hitting the /averages endpoint for the nodes-dashboard, there is no need to make that request again in the timeseries. For simplicity's sake, I just created a different timeseries comp that consumes the data that was already fetched in the nodes-deashboard and toggles the results in the tab

necoline:
great UI suggestions @rwegener2  Let's make them separate tasks to work on after the deploy

PR #563, opened 2021-01-13
Add dashboard card for mobile data locations

### LINKED ISSUE #551: Maps with Mobile data
## Goal
We not have been super concrete thus far about how we are displaying mobile data on maps.  This ticket documents the strategy for that.

### World Map / Countries Maps
* Show bounding boxes only, without a latest point
* (Nice to have) color the bounding box by the appropriate color.  Otherwise just leave it always as the black or whatever the default box color is

### Dashboards (Locations and Datasets)
**Nearby Locations component (Locations dashboards) and Datasets pages node selection**
* Same as world map

**Additional Map**
On only the mobile locations/datasets we should reintroduce a map to show the individual point locations.  This should be wired up to the date selection feature so when a user selects a particular date the point on the map change.

Styling is a nice to have right now.  If all the points show up in grey that is fine.  Interacting with the date selection is the higher priority.

### Note
Datasets I do not think are explicitly labelled as mobile or not.  We should consider a dataset to be mobile if any one of the locations it contains are mobile.  Right now I believe it will be all mobile or no mobile, but we want this to be flexible for future projects.

AliceR:
Closes #551 

This adds a card for mobile location pages where we could render all the measurement points. Since there could be a lot of points, we need to limit the tiles request in some way. 
@bitner mentioned zoom level and data range as possible limitations, I did not figure out the correct request yet to show all the points that I want to show up.

- [x] It would be nice to zoom the map to the extend of the bbox. For that it would be easiest for me if the bbox would be added to the /locations response - would that be easy enough for you to include, @bitner ?


jvntf (approved):
This looks good. @AliceR a couple of clarifications:
On the locations I checked I see one point and a rectangle, both gray. Is this the average case? But you are saying there could be a large number of points?
Not sure I understand why we need to limit the requests by zoom, aren't the tiles going to do that automatically? The world map for example displays a fairly large number of points without any issue.

necoline:
I don't get a bounding box for all locations but I do get at least one trace, is this expected?

AliceR:
@necoline @jvntf What I expect to see on a mobile location dashboard is a map with a bounding box, it should on load zoom to that bounding box, and further display the points where measurements were taken. It should look like a trace. For me it only shows the bbox and the last location (1 point), which is not what I expect. I am suspecting problems with the dateRange parameters send to the tiles request in the mobile-source - if you _do_ get a trace, for which location is that?

AliceR:
Turns our there is a bug in the backend, hence data is not yet showing up correctly. But, the tiles should generally work now. I disabled the location id filter for now so we can see other mobile points on zoom/pan.
I did connect the date range for the tiles filter to the collection dates, not yet to the daterange selector, but created a card for this remaining task (https://github.com/openaq/openaq.org/projects/1#card-52820685).

I would merge this for now and expect the points to show up later in the day, is that ok @rwegener2 ?

AliceR:
We decided to wait until backend is ready

AliceR:
Points are showing up for mobile dashboards:

Screenshot 2021-01-15 at 17 23 46

rwegener2:
I clicked quite a few pages.  The ones that worked looked absolutely beautiful.  I had some that seemed like they weren't working, but when I would go back to them a few minutes later they often would so I think it might be an API thing.

One clarification - Is that extra black line in this one just a nearby location's points?
http://localhost:3000/#/location/26351

Over all, let's merge it!  🎉

AliceR:
exactly, extra black line in this one just a nearby location's points. Though I am going to remove that with the tiles api's new capability of filtering by location - we have a limit of 1000 points returned and want to make sure to use that on the location that matters here.

PR #564, opened 2021-01-13
Improve site responsiveness

danielfdsilva:
This PR improves the responsiveness of the hub and single pages. This does not provide the best mobile experience but the presentation is no longer broken.
This required quite some changes:
Some of them were structural like for example ensuring that the html elements were grouped in a semantic way, instead of just looking correct in the page.
Other changes were just about the styles and the grid definitions for the different breakpoints.

For most of the style changes I relied on `sass` and `classes` since it had the theme variables and the media queries breakpoints and utility mixins already defined. Several of the styled components are also not coded to allow style composition.

I also had to remove the grid definition from the `card` components as this does not offer enough flexibility to implement media query breakpoints. Doing the following will no longer be possible:
```
<Card
  gridColumn={'1  / -1'}
```


AliceR (approved):

rwegener2 (approved):
this is great.  It really adds polish to not have the things overlap.

jvntf (changes requested):
Nice, do you mind adding a truncation or something for the card title in this PR 
image
Other than that looks great

AliceR:
@jvntf I created a [new card](https://github.com/openaq/openaq.org/projects/1#card-52753499) for your request in favor of this PR to be merged early (as soon as the check allows me to)

Issue #548, opened 2021-01-12
Update Content - Header description on search hubs


rwegener2:
## Goal
We currently have placeholder or outdated text in the headers of the Locations, Countries, and Datasets pages.  That should be updated with OpenAQ's new content.

### Where to add it
This text should go in the header of the Locations search, Countries search, and Datasets search.  Where I have {Location} in the text below can be substituted for Location, Country, or Dataset, respective to each page (without the curly brackets).

### Text to Add
Air Quality Data by {Location}

Our repository of air quality data continues to grow and now includes low cost sensor data in addition to reference-grade data. Learn more about the data types here. Explore this data through search features on the site or through the API, where all real-time and historical is now accessible.

We are currently collecting data in 99 different countries and primarily aggregate PM2.5, PM10, ozone (O3), sulfur dioxide (SO2), nitrogen dioxide (NO2), carbon monoxide (CO), and black carbon (BC) measurements. In certain limited cases, additional pollutants outside of those standard set of pollutants are now available through low cost sensor sources.

The low cost sensor addition to the platform is a pilot, let us know how you’re using it and give us feedback.
If you cannot find the location you are looking for, please [suggest a source](https://docs.google.com/forms/d/1Osi0hQN1-2aq8VGrAR337eYvwLCO5VhCa3nC_IK2_No/viewform) and [send us an email](mailto:info@openaq.org). 

[Data Disclaimer and More Information](https://medium.com/@openaq/where-does-openaq-data-come-from-a5cf9f3a5c85)

* Also be sure to include the "View API Docs" button/link from the existing website

<img width="955" alt="Screen Shot 2021-01-11 at 9 41 14 PM" src="https://user-images.githubusercontent.com/35503632/104266708-bde84000-5455-11eb-965e-898716a294f9.png">

PR #565, opened 2021-01-13
Update header content

danielfdsilva:
Added the content as outlines in #548.
I think this content is a bit too long to keep in the header. This is the result.
image


rwegener2 (changes requested):
Some things I noticed about the text (text in bold is what is missing from the headers):
* can we capitalize the "L" in "Air Quality data by location" to match the titles of the other pages?
> Our repository of air quality data continues to grow and now **includes** low cost sensor data ...

> **Learn more a**bout the data types here. (**Add spaces to the end of this sentence**)

> Explore this data through sea**rch features o**n the site or

> all real-time and historical **data** is now accessible.

Also just to confirm - is the country # (_We are not collecting data in 88 different countries ..._ ) dynamic?

Finally, I see what you are saying about the header being very long.  I'm not deeply concerned by it, but I wonder if we could increase the width of the text on the page a little or decrease the space above the title to make it a bit shorter.  You could make a decision on that yourself or reach out to @tania-pires.

danielfdsilva:
@rwegener2 wow. not sure what happened here. The country number is indeed dynamic.
I think the problem with the header is the amount of text, not so much the style of it. The user needs to scroll a bit before accessing the content, and this header is repeated in several pages. As long as this is ok, fine by me.

Review addressed and rebased from develop

rwegener2:
Thanks for the updates, @danielfdsilva!

I agree the length is long but I think it isn't high enough a priority to worry too much about right now.

The only place in which I will be a tad concerned is on the Datasets page.  We just got an additional blob of text (below) to add to this header to explain how the datasets page differs from the locations page.

> This page groups data from monitoring locations that form a sensor network where common sensor systems are managed by the same source and share standardized deployment practices, data post-processing and quality assurance, that has been documented in related metadata. Datasets allow for additional exploration of air quality across comparable locations within the same network. Navigate to a specific dataset’s page to view measurements and find technical information on collection methods. Datasets are part of a pilot to incorporate low cost sensor data into the platform. Let us know how you are using datasets and give us feedback.

Adding this, I think, would be way too long.  @danielfdsilva do you have any ideas of something low lift to get this in the header without extending it too much?  I was thinking about hover text, either over the words "by Dataset" in the title or adding a fun bit of text below the "View API Docs" button that says "What's a dataset?" and adding the hover/click text there.  If you have another idea that would be easier for your to implement I'd also be all for it.

rwegener2:
We have the confirmed text:
> Our repository of air quality now includes low cost sensor data as a pilot, in addition to reference-grade data. Learn more about the data types here. Explore this data through the site or through the API, where all real-time and historical data is now accessible.

> We are currently collecting data in 99 different countries and primarily aggregate PM2.5, PM10, ozone (O3), sulfur dioxide (SO2), nitrogen dioxide (NO2), carbon monoxide (CO), and black carbon (BC) measurements. Additional pollutants outside of those standard set of pollutants are now available through low cost sensor sources at certain locations. If you cannot find the location you are looking for, please [suggest a source](https://docs.google.com/forms/d/1Osi0hQN1-2aq8VGrAR337eYvwLCO5VhCa3nC_IK2_No/viewform) and [send us an email](mailto:info@openaq.org).

[Share Feedback](https://docs.google.com/forms/d/1s5GYME3He-EDed-jgg11Ry3zmvTZkPK5xSnQUWwu0Gk/edit) (Button)
[Data disclaimer and more information](https://medium.com/@openaq/where-does-openaq-data-come-from-a5cf9f3a5c85)

That should be the text on the Locations, Countries, and Datasets hubs.  The Datasets hub will be this plus the paragraph pasted above [in this comment](https://github.com/openaq/openaq.org/pull/565#issuecomment-759899322).  It will be a lot of text but it is what we are doing for now.

danielfdsilva:
@rwegener2 content updated

rwegener2 (approved):
This looks nice.  I like what you did with the header making "Air Quality Data by" smaller.

Thanks, @danielfdsilva!

PR #568, opened 2021-01-13
Update temp coverage card

### LINKED ISSUE #558: Updating Temporal Coverage
## Goal
The Temporal Coverage component often shows an error.  To address these errors we want to change when each piece of the the component is displayed.

## Details
The temporal coverage has three sections - hour of day, day of week, and month of year.  Which section is displayed when is what is changing.
✅   means display that component
### Changes to how the component is displayed mobile vs. stationary
|   | Hour of Day | Day of Week  | Month of Year |
|---|:---:|:---:|:---:|
|  Mobile Location or Dataset |  ✅   | ✅    |✅   |
|  Stationary Location or Dataset | remove  |  ✅   | ✅  |

### Changes across date selection
|   | Hour of Day | Day of Week  | Month of Year |
|---|:---:|:---:|:---:|
| Full Project Lifetime |  ✅  (mobile only) | ✅    |✅   |
|  One Month of Data | ✅   (mobile only)  |  ✅   | ✅  |
|  One Day of Data | ✅  (mobile only)  |  ✅   | message like "not applicable" |

## Note
If any single location in a dataset is mobile the whole dataset is considered to be mobile

jvntf:
closes #558


rwegener2:
This is sort of half of a review, because at the time I was reviewing the locations dashboards were crashing.  I was able to use a dashboard page to notice a couple things.
## Example project http://localhost:3000/#/project/27585
Sub headings refer to the date selection
### Lifetime of project
* This may be done in the locations dashboards but I would like to remove Hour of Day.  It is never applicable to a stationary page.  Ideally the remaining two charts would be centered, I think, but given that styling is hard whatever happens by default is probably fine.
### One Month
http://localhost:3000/#/project/27585?dateRange=2019/6
* Looks like the month is showing an empty chart (example below).  I think ideally it would show just one bar with the single month of selection and the number of measurements on the y axis.  If this isn't compatible with the API response then I would write "Month of Year chart not shown for a one month of date range selection"
Screen Shot 2021-01-13 at 10 07 42 PM
### One Day
* If something is returned from the API I'd roll back the original instruction in the ticket and have month of year show just the single month bar.  If that isn't compatible with the API I'd go with a more specific error message: "Month of Year chart not shown for a one day of date range selection"

Mobile Page
http://localhost:3000/#/location/60749 (errors at time of viewing, time series and temporal coverage charts are empty)

These are a few comments that I think will be applicable to the locations pages.  I'll check back in tomorrow to see if the pages are loading again and give more another review.

jvntf:
@rwegener2 issue with month chart addressed. Projects endpoint is not providing `isMobile`. Removed charts when there is no data

rwegener2 (changes requested):
Something is off about the mobile data but that has been true across PRs so I think that is something to address once we have things more broadly merged.

Just looking at stationary locations this looks good.  Removing charts with no data was a good idea - it does the job quickly and saves us from the complexity of different error messages in different scenarios.

I did notice that there is a lingering "(" hanging out where the Hour of Day chart used to be.  This was true I think on all the pages I looked at, but http://localhost:3000/#/location/60760 is one example
Screen Shot 2021-01-14 at 11 40 52 AM

Summary - the functionality looks great.  Please remove the "(" if possible

Issue #543, opened 2021-01-11
Countries Hub updates


rwegener2:
## Goal
Walk through the countries hub and make sure that the existing OpenAQ functionality is in alignment with our updates and the presence of LCS data.

## Details
Checkboxes are things to check on.  Quotes below come from the feedback documents.

- [x] ~Investigate what is meant by number of locations~ There is only one definition of a location, the one we have been using throughout this project.  So the number of locations number is incorrect.

Ex. Germany & Canada have lots more locations on their respective pages than are listed on the card; Costa Rica doesn't seem to have any even though the card says 1.
> Australia lists 13 but there are many more on the country page.
- [x] Investigate why nothing shows up next to measurements (countries cards and countries page headers)
- [ ] Some of the countries are missing (Ex. Andorra, Bahrain, Bermuda)

### Nice to Have
- [x] Scrolling
> When scrolling through the country pages, the mouse gets stuck on the map that zooms in and out. Is there a way to fix this so that it’s easier to scroll through the page?

AliceR:
> Investigate why nothing shows up next to measurements

We need to add the measurement count to the /v2/countries/:id response, added a card [here](https://github.com/openaq/openaq.org/projects/1#card-52733552) (cc @bitner)

AliceR:
> the number of locations number is incorrect.

the number displayed is provided with the /v2/countries/:id response as `locations`

AliceR:
While investigating the above, I noticed that the label `areas` never shows a number - it is currently set up to display what is provided with the /v2/countries/:id response as `cities`, but I could not find an example where this exists.

For the number of sources, we currently iterate over the locations cards and find unique values. I think it would be better to include that to the endpoint.

AliceR:
> Some of the countries are missing (Ex. Andorra, Bahrain, Bermuda)

Those are not provided with /v2/countries

rwegener2:
Also:
add number of sources to /v2/countries objects

rwegener2:
What you said about areas returning cities makes sense @AliceR I think we should remove the `areas` number because the majority of the data now does not have a city, or at least not a valid one

rwegener2:
### Frontend actions
All of the backend issues raised here have been addressed, with just one left in progress (A few missing countries, see note below).  That leaves the following frontend tasks re-open for development:
* updates sources to use the endpoint directly instead of iterating
* remove areas from the countries page

### Extra background
Outcome of the missing countries problem:
[...]

PR #569, opened 2021-01-13
Small fixes for country page

AliceR:
Contribute to #543 

- Filter needs to be updated to be compatible with ["all", ...]
- Skip areas label in country page header when the number is not available
- Disable scroll zoom on country map (It is still possible to zoom with buttons!)


rwegener2 (approved):
I'm not sure I understood what this comment means:
> Filter needs to be updated to be compatible with ["all", ...]

However the measurements numbers looks good and the disabled scroll is really nice

PR #577, opened 2021-01-13
Fix/576 hourly

### LINKED ISSUE #575: Dashboards - Single date selection only shows up on refresh

### LINKED ISSUE #576: display hourly data instead of daily data when 1 month date selection is chosen
## Where
Dashboards time series component

## Details
When the user has selected a single month in the date selection feature instead of showing one point per day the chart should show one point per hour.  This should be allowable by the API

jvntf:
closes #576 

closes #575 

branches off of #574

### DIFF
--- a/app/assets/scripts/components/dashboard/time-series-card.js
+++ b/app/assets/scripts/components/dashboard/time-series-card.js
@@ -78,7 +78,13 @@ export default function TimeSeriesCard({
       } else if (projectId) {
         query = { ...query, project: projectId, spatial: 'project' };
       }
-
+      //If date range is not lifetime, get hourly data
+      if (dateRange) {
+        query = {
+          ...query,
+          temporal: 'hour'
+        }
+      }
       fetch(
         `${config.api}/averages?${qs.stringify(query, { skipNulls: true })}`
       )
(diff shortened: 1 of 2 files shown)


rwegener2:
I wasn't able to get a review of this one because, at least the time of this post, I couldn't get any data to render.  This seemed to be connected to the date selection since all the pages I tried were able to render the default time series for the project lifetime.  Sometimes data I think not rendering means the database is going up or down.  In the morning @jvntf if this is working on your computer I can try again.

### Example 1
* http://localhost:3000/#/location/301?dateRange=2020/10/13 or http://localhost:3000/#/location/28692?dateRange=2020/12

On this page I see an empty graph (what I mean by "empty graph" is below).  Is there no data?  If so we should render a message like:
> No data is available in your selected time window.  You may have chosen a date outside of the collection date range or a date within the collection date range in which the sensor was not reporting"

### Example 2
* http://localhost:3000/#/location/22?dateRange=2015/2

This one was interesting because if you are in Feb 2015 it shows the "We couldn't find any data." message but March 2015 is an empty graph.  All of the individual days I tried resulted in the "We couldn't get any data" message.

### "Empty graph"
Screen Shot 2021-01-13 at 8 18 52 PM

jvntf:
@rwegener2 ok this should be good to go now (hopefully)

rwegener2 (approved):
Looks great!  Both issues seem to be well resolved.

If it takes 10 min or less it would be nice to add the timestamp to the hover text on the time series.  If it would be more than a 10 min process, though, we can live without it.

Issue #571, opened 2021-01-13
bug - make sure you can click on a location on the world map and be taken to the dashboard (nice to have - add latest value to popup)


jvntf:

jvntf:
@rwegener2 i cannot replicate any bug here. can you confirm and / or close?

rwegener2:
@jvntf you are right the popup does appear to be working on the frontend.

To rebrand this ticket a bit then I'd say the task would be to make sure that when you click on a mobile square a pop up still shows up (Ex. ozone measurements in southern Austria)

rwegener2:
Actually it looks like some of the mobile measurements are showing pop ups, but the locations themselves are shown as square points.  Maybe wait on this ticket until the styling of the visualization of mobile data is confirmed with @AliceR

rwegener2:
A couple of things I noticed about the popups:
* The source doesn't seem to link to anything.  We could add the source link or for simplicity I would totally support just listing the source as text.
* "View More" is a bit generic - we could change this to "View Location" or "View Dashboard"

AliceR:
These should all be tackled in #579

PR #579, opened 2021-01-14
Update maps

AliceR:
- Grey out measurements older than 2 days
- Update filter criteria for measurement symbols (sensorType)
- Fix last reading on popover, it should now display eg. "Last reading 72.00 µg/m³ at 2021/01/13 22:00"
- Fix source url on popover
- Update button text on popover
- Add popover to mobile bounds layer

### DIFF
--- a/app/assets/scripts/components/map/legend.js
+++ b/app/assets/scripts/components/map/legend.js
@@ -140,7 +140,7 @@ export default function Legend({
               </li>
             ))}
           </ul>
-          <p>* Locations not updated in the last week are shown in grey.</p>
+          <p>* Locations not updated in the last two days are shown in grey.</p>
           <small className="disclaimer">
             <a href="https://medium.com/@openaq/where-does-openaq-data-come-from-a5cf9f3a5c85">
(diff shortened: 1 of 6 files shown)


rwegener2 (approved):
This looks great 👍 

Popover reading and adds a lot to the interactivity of the map, all the locations I clicked on have accurate colors based on their reading, and the square points are all LCS stations.

I did notice that popups do not exist on the mobile datasets.  The mobile datasets I found are showing up as boxes (🎉 ), they just can't be clicked (Ex. a group of boxes in southern Bristol, UK on PM2.5).  Ideally the point would also get removed from the mobile box, but that is less important than the popup.  This may already be getting worked on so @AliceR you should decide if that is happening here or in another PR.

rwegener2:
Just saw the comment in #571 about some issues being addressed here.  Looking at that issue:

* source links are working well
* I like the "View Location" text

I can revisit again later to review mobile popups.

AliceR:
bbox popups should have been included with 5a6a957 - can you check again, please?

Issue #562, opened 2021-01-13
Data download buttons


rwegener2:
## Goal
Fully wire up download data buttons

### Where the buttons are
* Locations hub cards & Location dashboard header
* Dataset hub cards & Dataset dashboard header
* Countries cards & Countries page header

### Functionality
- [ ] Show the parameters with their `displayName` instead of their `name`
- [ ] Group the 7 OpenAQ Core filters at the top (find these at https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/parameters `results` -> _index_ -> `isCore`).  If we can add just a line or something to separate those from the parameters below that would be great.  The 7 could be labelled "Core Parameters" and the others could be labelled "Additional Parameters"
- [ ] Make sure all locations of the button above direct to the download popup (Ex. countries cards download button links to the homepage)
- [ ] "Download selection" button should generate a csv

**Nice to have**
- [ ] Auto-populate Location if you clicked the Download button on a location page or card

PR #580, opened 2021-01-14
Donwload modal

danielfdsilva:
This PR re-implements the Download modal for locations.

- Migrate download modal code to functional components
- Show the parameters with their `displayName` and group them by core/non core
- Ensure Download buttons open the modal (done for locations and country single and hub pages)
- Location fields pre-selected depending on context.

image

### Questions
The modal header displays a message about data availability. Is this still correct?
The `Realtime -> https://openaq-fetches.s3.amazonaws.com/index.html` links to something that doesn't work.
The `daily -> http://openaq-data.s3.amazonaws.com/index.html` looks very outdated.

Everything should be connected and once the api in in place it should work (or minimal changes are required)

### Blockers

The download is not implemented for the datasets, because they're a different entity entirely.
A location has a country/city/location hierarchy (Most actually, some are under "No listed city"). The datasets seem to only have a country, and by looking at the response, the `countries` property is an array. Is it correct to assume a dataset can have multiple countries? Can a dataset not have a country at all?

Besides the above, the endpoint to get a list of datasets is different and the current modal structure can't be used.
A possible option is to add a radio button to the modal where you select what you want to download data for (dataset or location), and change the fields based off of that.

Two options moving forward:
- Ensure the modal works with locations and mark the datasets as coming soon
- Implement the option above


danielfdsilva:
Added a commit to temporarily disable the data download

image

rwegener2:
closes #562

rwegener2:
@danielfdsilva The measurements endpoint should now properly allow the user to download a csv

### Example & Details
https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/measurements?format=csv
It should be able to be filtered by location.  For the datasets page David recommended using the same endpoint but filtering by the list of locations.  Since a dataset is just a list of locations filtering by a dataset should just mean filtering by all of the locations associated with that datasets.  If this doesn't make sense feel free to ping.

danielfdsilva:
Datasets added as download option.
image

Still left to confirm the header text.

rwegener2:
This looks fantastic.

I say replace the header text with:
> Customize the data you want to download, now including [low cost sensor data](link to blog post). For greater customization, use the [API](http://docs.openaq.org). [Realtime](https://openaq-fetches.s3.amazonaws.com/index.html) archives of all 841,689,793 measurements are also available. 
[Share your feedback](https://docs.google.com/forms/d/1s5GYME3He-EDed-jgg11Ry3zmvTZkPK5xSnQUWwu0Gk/edit) on the low cost sensor pilot so we can continue to improve the platform!

And then merge it!

rwegener2 (approved):

PR #581, opened 2021-01-14
Fix card title overflow

danielfdsilva:
From: https://github.com/openaq/openaq.org/pull/564#pullrequestreview-567272266

Result:
image

This also includes a fix for some filters bugs that happened when loading the page with filters selected (through the url)


rwegener2 (approved):
This looks great to me.  Thanks for addressing some filtering issues as well, Daniel.

If anyone else is looking for some cards with this issue I went to the locations page and filtered on the parameter "PM2.5 count" and several cards on the first page had this change.

PR #582, opened 2021-01-14
Fix/country sources

necoline:
removes undefined sources in country page
adds catch to missing sources in location card


AliceR (approved):
This fixes the issue I experienced with some country pages.

More in general, we should probably add the number of sources to display in the header to the /v2/countries/:id response and avoid the frontend data wrangling.

jvntf (approved):

PR #585, opened 2021-01-14
add source seperator

### LINKED ISSUE #550: Bug - Distinguish between sources on results cards if multiple sources
In the example below 'US EPA AirNow' should be more easily distinguished from "Albuquerque Environmental Health Department".  A new line or a semicolon or another solution would be good. 

Screen Shot 2021-01-11 at 9 37 37 PM

This example image came from http://localhost:3000/#/locations?sources=albuquerque-environmental-health-department

jvntf:
closes #550

### DIFF
--- a/app/assets/scripts/views/locations-hub/location-card.js
+++ b/app/assets/scripts/views/locations-hub/location-card.js
@@ -59,9 +59,10 @@ export default function LocationCard({
               label: sources?.length > 1 ? 'Sources' : 'Source',
               value: sources?.map((source, i) => (
                 <a
-                  href={source.sourceURL}
+                  href={source.url}
                   title={`View source for ${name}`}
                   key={source.name}
+                  className={!source.url && 'disabled'}
                 >
                   {`${i > 0 ? ', ' : ''}${source.name}`}
                 </a>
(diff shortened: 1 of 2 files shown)


necoline (approved):

AliceR (approved):

rwegener2 (changes requested):
The comma separator took me a second to find.  That makes me want to switch to a new line but I think then some of the cards would get really long.  So long story short I think the comma solution is good.

While this PR is open can we also switch to only show the source name in blue if there is actually a link?  If this is a lift then I think showing the source name in black is fine.  
If anyone else has ideas how the coloring of the links should behave I'm all about it (@AliceR did you just update a source link for a PR?).

For anyone looking for some examples of the fix I found a few here http://localhost:3000/#/locations?sources=albuquerque-environmental-health-department&page=853

AliceR:
Daniel did some changes to the way sources are displayed on the dashboard card here: https://github.com/openaq/openaq.org/commit/cee42efdb6003f92c87056af2119fd6c029768f8#diff-565884f2f6132bd49632635af45977014e59507b5d4b1e9f57b22c8d2dfa2297R33

PR #590, opened 2021-01-15
Fix: remove basedata

AliceR:
Currently, the whole page crashes if just one endpoint fails (right now, `/countries`). This PR changes removes the request for "base data" on every page, and fetches this data only when it is actually needed.

I am sure there is more refactoring that we can do to further improve the code, but for now it fixes the broken staging link.


rwegener2:
From a user side this works great for me -- all the pages except the broken countries page render 🎉

app/assets/scripts/views/locations-hub/locations-hub.js
```diff
@@ -94,10 +92,8 @@ export default function LocationsHub({
 
       <div className="inpage__body">
         <Filter
-          parameters={parameters}
-          countries={countries}
-          sources={sources}
-          manufacturers={manufacturers}
+          slug="/locations"
+          by={['parameters', 'countries', 'sources', 'sensor']}
```
jvntf, line 96: why `by`? just curious what that means

jvntf (changes requested):
I think a few other components need to `fetchBaseData`
At least:
`views/country/`
`views/compare`

I believe if a user navigates `locationsHub` -> `location` -> `compare`, the baseData will be in the state, but if they navigate directly to `compare` the data is not there.
`/country/{id}` seems to always fail due to `parameters` prop being undefined.

Thank you for taking on these big refactor upgrades!!!

app/assets/scripts/views/locations-hub/locations-hub.js
jvntf: no that works!

jvntf:
@AliceR sorry submitted review before your last commit

app/assets/scripts/components/filter.js
```diff
@@ -156,108 +167,117 @@ export default function Filter({
...
-                <ParamSelect
-                  parameters={parameters}
-                  onFilterSelect={onFilterSelect}
-                  selected={selected}
-                />
-              </Dropdown>
+              {parameters && parameters.length > 1 && by.includes('countries') && (
+                <Dropdown
+                  triggerElement="a"
+                  triggerTitle="View filter options"
```
jvntf, line 209: Tests want this to say `View parameter options`, not sure why this wasn't failing previously

necoline (approved):

jvntf (approved):
lgtm!

PR #593, opened 2021-01-15
Add entity tag

### LINKED ISSUE #592: Add entity tag to locations results cards (just like sensorType or isMobile tags)

jvntf:
closes #592


rwegener2 (changes requested):
Looks pretty good.  A few comments:
* The tag on my screen overflows the card.  I know styling has been a bit of a pain, might we put the isMobile tag on a separate line as a less complicated fix?
Screen Shot 2021-01-15 at 2 51 26 PM
* I'm just noticing now that when I hover over the tag my cursor turns to imply that I can click the tag like a link.  Can we stop that from happening, since there is no interaction on the tag?

jvntf:
@rwegener2 feedback addressed

rwegener2 (approved):
I like it.  If it's easy to center them all I wonder if that would make it look a bit nicer?  But I don't really know 🤷‍♀️.  Other reviewers could weigh in.

Either way @jvntf play with it if you want but I think I this is ready to merge.

rwegener2:
I was playing with the site and noticed something else - 

@jvntf Can we also add an entity tag to the header of the dashboard page?  And if you could do the same thing with removing the click hover that would be amazing.

Thank you thank you!

AliceR (approved):
I added the entity tag to the location dashboard in 9af8800 and changed from grid to flex, which helps with the spacing on line break.

rwegener2:
Looks great, thanks @AliceR!

PR #594, opened 2021-01-15
Update/552 ppb

### LINKED ISSUE #552: Display concentration units in ppb and number rounding
## Goal
Any units that the database reports in ppm we would like to be displayed as ppb on the frontend.  These numbers should be rounded to the nearest whole number.

### Where
This change will be applicable to the following parameters: CH₄, CO, CO₂, NO, NO₂, O₃, SO₂, NOx
At least the following components will be affected: Latest measurements; Measurands, Time Series (including the hover)

### How
ppm * 1000 = ppb
So a measurement of 0.25 ppm = 250 ppb (0.25 * 1000 = 250)

Once the new number is calculated it should be rounded to the nearest whole number.

jvntf:
closes #552 

- [x] Convert units to ppb if ppm
- [x] render ppm as ppb
- [ ] Apply this to map and map legend


jvntf:
I did not apply this to the map because it would affect some of the logic that is currently hardcoded, which i believe is going to be removed soon?

AliceR (approved):

PR #600, opened 2021-01-18
Update home: add back stats section

AliceR:
A bit of refactoring to split the home view into manageable file sizes.

To display the **total number** of `locations`, `sources`, `countries`, and `measurements`, we previously fetched all those endpoints and stored their results in the base data/base stats redux store. With the new api setup, it would make more sense to get these statistics from a dedicated endpoint (cc @bitner) .


rwegener2:
Thanks for following up on this @AliceR and starting the PR.  I added this to David's API requests sheet.

necoline (approved):

PR #602, opened 2021-01-18
updates api addresses

necoline:


app/assets/scripts/config/production.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     token: 'pk.eyJ1IjoiZGV2c2VlZCIsImEiOiJnUi1mbkVvIn0.018aLhX0Mb0tdtaT2QNe2Q',
     baseStyle: 'mapbox://styles/devseed/ciqs29d060000clnr9222bg5x',
   },
-  api: 'https://api.openaq.org/v2',
+  api: 'https://u50g7n0cbj.execute-api.us-east-1.amazonaws.com/v2',
```
AliceR, line 13: I think this should actually stay as is
```suggestion
  api: 'https://api.openaq.org/v2',
```

AliceR (changes requested):

app/assets/scripts/config/production.js
necoline: cool, I wasn't sure. I could read bitner's message both ways

AliceR (approved):

PR #614, opened 2021-01-18
Update deploy.yml

sruti:


necoline (approved):
I'm sure we can look into a more elegant way to do this down the road, but this should do for now

PR #617, opened 2021-01-18
Update deploy-prod.yml

AliceR:
Since we skip the QA process, it seems we need to skip this conditional as well


sruti (approved):
Good catch! I think QA may still run, but Prod Deploy won't wait till it's done to run. To fix it, we can have lint-and-test skip master, then add those jobs to prod deploy. But let's try this for now!

PR #619, opened 2021-01-19
Reuse source list for popover

AliceR:
This makes the non-clickable sources on the popover appear non-clickable, the same way as in the dashboard card.

### DIFF
--- a/app/assets/scripts/components/dashboard/sources-card.js
+++ b/app/assets/scripts/components/dashboard/sources-card.js
@@ -4,16 +4,7 @@ import styled from 'styled-components';
 
 import SmartLink from '../smart-link';
 import Card from '../card';
-import config from '../../config';
-
-const SourceList = styled.ul`
-  padding: 0;
-  list-style: none;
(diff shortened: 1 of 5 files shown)


rwegener2:
Is this intended to fix apply to the pop overs on the world map?  If so the links that do exist work nicely (and I didn't think they were working before), but I still see a link and hover on some of the LCS sensors that don't have links.  For example:
Screen Shot 2021-01-19 at 7 59 09 AM
Screen Shot 2021-01-19 at 7 59 28 AM

While this PR is open it would be awesome if we got the sources that do have links to open the link in a new tab.  We can still keep the unified component, I think the source info should open in a new tab both on the world map and on the dashboard cards.

AliceR:
Excuse me @rwegener2 , I forgot to commit the new file. Updated the branch, and also included the organizations to the project card.
Waiting for my pushed commits to show up...

rwegener2 (approved):
Looks great 🎉

PR #622, opened 2021-01-19
Add/572 datasets filters

### LINKED ISSUE #572: Additional filters to dataset search
## Goal
We would like to add the following filters to the datasets search:
* Sensor Type filter
* add Data Source filter

## Details
I think this is possible using the `/projects` endpoint.  The data source and sensor type filters should look like those same filters on the [locations page](https://openaq.org/#/locations).  The difference with datasets is that the projects are tagged with a list instead of a single string so we should return results if they include that item in their list.

~For example, a dataset with the tags 'community' and 'government' would get returned if these entities were chosen:~
* ~just community~
* ~government entity~
* ~both community & government~

~It would not get returned if:~
* ~just research~
* ~both research & community~
* ~both research & government~
* ~all three research, community & government~

### Update 
I made a mistake - each dataset gets assigned one value of the predominant sensor Type.  I think this filter should just work like the locations, in that case, returning all of the locations that match the selected value.

It's worth noting that right now none of these complex scenarios exist because none of our current datasets should have multiple tags.

jvntf:
closes #572 
Also adds tags to project cards 

Sensor type query is not working currently.  projects endpoint seems to expect boolean values for entity and sensorType as opposed to locations which takes string. Needs clarification re: is this a bug or not

### DIFF
--- a/app/assets/scripts/views/projects-hub/project-card.js
+++ b/app/assets/scripts/views/projects-hub/project-card.js
@@ -17,6 +17,8 @@ export default function ProjectCard({
   totalMeasurements,
   parametersList,
   sources,
+  sensorType, entity,
+  mobile
 }) {
   let updated = moment(lastUpdated).fromNow();
   let started = moment(firstUpdated).format('YYYY/MM/DD');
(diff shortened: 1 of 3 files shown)


AliceR:
Something seems wrong with the sensor type filter, selecting "Reference" still shows the low cost sensor projects...?

jvntf:
@AliceR hm i think that is on api side, i will ask bitner

jvntf:
enabled data source filter here as well, though this is not yet supported by the api

jvntf:
Everything here now working correctly with api updates. cc @rwegener2 @AliceR

rwegener2 (approved):
This looks great to me.  Thanks for working through all the nitty gritty details @jvntf

PR #628, opened 2021-01-20
Update how mobile data is displayed

### LINKED ISSUE #621: add mobile map to the datasets page & update locations mobile map
## Goal
Display a map with the more details points of mobile locations in a dataset similar to how we do on the locations page.  These changes apply only to locations or datasets where the isMobile field is True.

## Details
The maps should both use a new endpoint to display mobile data:
> Changes just deployed to staging:
Created new endpoint /v2/locations/tiles/mobile-generalized/{z}/{x}/{y}.pbf
"default" layer has the generalized/grouped points which only has the locationId
"bounds" has the bounding box which has all the same attributes as the single static points
I removed the "bounds" layer from the /v2/locations/tiles/{z}/{x}/{y}.pbf
The actual measurement data filtered by time for mobile locations is still at /v2/locations/tiles/mobile/{z}/{x}/{y}.pbf, but I'm not sure if we still want to use that or if it would just be better to use the mobile-generalized endpoint
These are all pushed up to staging and the viewer to see working sample is up at https://b4qc1cqcm2.execute-api.us-east-1.amazonaws.com/v2/locations/tiles/viewer

This should display all of the points of the location instead of the one that previously just had a few weeks of info.
For the dataset this should display all of the locations in that dataset, up to 15 nodes.  Greater than 15 locations I think it should say something like "map view not available with greater than 15 nodes".  This shouldn't be a warning that gets used with the current data, since all of the mobile datasets have fewer than 15 locations right now.

### LINKED ISSUE #627: datasets page map styling
### 1. Stationary datasets
The map on the stationary datasets shows their points being circles when they should be squares for LCS data
Screen Shot 2021-01-20 at 1 06 29 AM

Example datasets (complete list):
https://openaq-staging.org/#/project/28988 (screenshot above from this dataset)
https://openaq-staging.org/#/project/29097
https://openaq-staging.org/#/project/29197

These do show up nicely as squares in the world map:
Screen Shot 2021-01-20 at 1 05 57 AM

### 2. Remove node selection from the mobile datasets
It was decided that the node selection is not relevant to mobile datasets, so it can be removed from these pages.  Date selection should stay.  The map view will be replaced by the one created in #621

AliceR:
Close #621 
and
Close #627

### DIFF
--- a/app/assets/scripts/components/dashboard/mobile-data-locations-card.js
+++ b/app/assets/scripts/components/dashboard/mobile-data-locations-card.js
@@ -4,7 +4,6 @@ import styled from 'styled-components';
 
 import Card, { CardHeader as BaseHeader, CardTitle } from '../card';
 import Map from '../map';
-import LocationsSource from '../map/locations-source';
 import MobileSource from '../map/mobile-source';
 import MobileBoundsLayer from '../map/mobile-bounds-layer';
 import MobilePointsLayer from '../map/mobile-points-layer';
@@ -30,14 +29,12 @@ export default function MobileDataLocationsCard({
       )}
       renderBody={() => (
         <Map bbox={bbox}>
-          <LocationsSource>
-            <MobileBoundsLayer locationId={locationId} />
-          </LocationsSource>
           <MobileSource
             locationId={locationId}
             firstUpdated={firstUpdated}
             lastUpdated={lastUpdated}
           >
+            <MobileBoundsLayer locationId={locationId} />
             <MobilePointsLayer locationId={locationId} />
           </MobileSource>
         </Map>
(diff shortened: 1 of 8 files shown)


rwegener2 (approved):
This looks quite nice.

I noticed on some of the larger projects (Houston Mobile (http://localhost:3000/#/project/28972) or Breathe London Mobile (http://localhost:3000/#/project/28979)) it is taking a minute for the data to show up.  Could we add a loading message while this is happening?

Breathe London Mobile and Long Island Mobile Methane (http://localhost:3000/#/project/28961) I noticed have one leg of the bounding box that get cut off
Screen Shot 2021-01-20 at 6 33 54 AM
This isn't ideal, but it also isn't the worst thing right now so I'm flagging it for investigation either now or later.

The other future thing is that now that these are a count of measurements either a popover to click or a color bar by number of count would be great.  This can also be saved for an iterative PR, however.

PR #630, opened 2021-01-20
Specify environment in deploy workflow

AliceR:
Our current production environment is not pointing to the expected endpoints. It seems to be an issue related with the config setup, see [here](https://github.com/openaq/openaq.org/blob/develop/gulpfile.js#L43). I am assuming we don't have the TRAVIS_BRANCH env var anymore after switching to github actions, and setting the DS_ENV explicitly will fix the issue. @danielfdsilva can you confirm this would work as I expect?

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -4,21 +4,25 @@ name: Production Deploy
 
 on:
   push:
-    branches: 
+    branches:
       - master
 
 jobs:
   build-and-deploy:
-    runs-on: ubuntu-latest 
+    runs-on: ubuntu-latest
+
+    env:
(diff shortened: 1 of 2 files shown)


danielfdsilva:
@AliceR yes. setting the variable manually will always take precedence and will fix the issue

jvntf (approved):
This makes sense. Perhaps we should remove this from [gulpfile](https://github.com/openaq/openaq.org/blob/4c4a6f950f9567f6c95420ab2a817c8178d16d9e/gulpfile.js#L44) as well? as far as i can tell, TRAVIS_BRANCH and DEPLOY_BRANCH are no longer set anywhere, maybe makes sense to just default to `DS_ENV=production` if variable is not set.

PR #636, opened 2021-01-20
Add loading animation to map

AliceR:
This shows a loading animation while the tiles are loading - on all maps.
At times it disappears before the layers show up, that seems to be related to slow rendering? Would need to investigate a bit more...

### DIFF
--- a/app/assets/scripts/components/map/map.js
+++ b/app/assets/scripts/components/map/map.js
@@ -3,11 +3,13 @@ import PropTypes from 'prop-types';
 import mapbox from 'mapbox-gl';
 
 import config from '../../config';
+import LoadingMessage from '../loading-message';
 
 export default function Map({ center, bbox, scrollZoomDisabled, children }) {
   const containerRef = useRef();
 
   const [map, setMap] = useState(null);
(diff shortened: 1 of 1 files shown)


jvntf (changes requested):
Im seeing some situations in which the loading message is not updated even though `areTilesLoaded` is returning true. I'm trying to understand why this is happening / find a way to reproduce but lmk if you don't see the same. I tend to see it happen after changing params  a few times, seems to appear on CO for whatever reason. @AliceR

jvntf:
image
That `true` is getting logged like: 
```
 useEffect(() => {
    setInterval(() => {
      if (map) console.log(map.areTilesLoaded())
  }, 200)
  }, [map])
```

jvntf:
starting to wonder if this feature is necessary. If we are to use `areTilesLoaded` to toggle a loading graphic, this means the graphic shows up very frequently. I.e. panning / zooming the map -> any time we request new tiles the graphic should appear at least for a split second. As of now, I am not noticing any significant lag in loading the layers.

AliceR:
You are right @jvntf  - especially with the new server side caching we might not need this at all. Closing now, we can come back to it when we really need to.

Issue #626, opened 2021-01-20
bug - download data some locations not pre-populating


rwegener2:
Sometimes when you click "download on some of the locations the Country/Area/Location does not pre-populate as expected.  the modal just hangs waiting for you to select a country:

<img width="767" alt="Screen Shot 2021-01-20 at 12 51 29 AM" src="https://user-images.githubusercontent.com/35503632/105137868-a34b4200-5ab9-11eb-8e87-bebdeba55085.png">

Our partners reported this issue although it seems pretty uncommon.  This is the only location I found where it was happening:
https://openaq-staging.org/#/location/67437

Issue #625, opened 2021-01-20
bug - download modal not filtering results


rwegener2:
When I click on download for a particular location or dataset the resulting csv gives me many more locations than just the ones that are part of the location that I clicked on. 
For example, the download button from the following location (https://openaq-staging.org/#/location/62446):
<img width="428" alt="Screen Shot 2021-01-20 at 12 38 20 AM" src="https://user-images.githubusercontent.com/35503632/105136748-cd036980-5ab7-11eb-8a05-a0e639da5462.png">
returned a csv like:
```
location,city,country,utc,local,parameter,value,unit,latitude,longitude
Spring Valley Township,,US,2021-01-19T18:41:29+00:00,2021-01-19T17:41:29-01:00,pm1,6.8,µg/m³,39.6005,-83.9736
University of Patras,,GR,2021-01-19T18:41:29+00:00,2021-01-19T17:41:29-01:00,pm1,17,µg/m³,38.2894,21.7855
Spring Valley Township,,US,2021-01-19T18:41:29+00:00,2021-01-19T17:41:29-01:00,pm25,11.4,µg/m³,39.6005,-83.9736
University of Patras,,GR,2021-01-19T18:41:29+00:00,2021-01-19T17:41:29-01:00,pm10,23.9,µg/m³,38.2894,21.7855
Spring Valley Township,,US,2021-01-19T18:41:29+00:00,2021-01-19T17:41:29-01:00,pm10,13.3,µg/m³,39.6005,-83.9736
University of Patras,,GR,2021-01-19T18:41:29+00:00,2021-01-19T17:41:29-01:00,pm25,21.8,µg/m³,38.2894,21.7855
The Hub,,,2021-01-19T18:41:28+00:00,2021-01-19T10:41:28-08:00,pm10,11.1,µg/m³,48.7297,-123.6641
Kingston,,AU,2021-01-19T18:41:28+00:00,2021-01-20T05:41:28+11:00,pm1,0.8,µg/m³,-42.9692,147.2652
```
Notice that the `location` column contains many more results than just "Spring Valley Township".

Can we please investigate where the error is being introduced?  Are we sending the correct fields to the backend or is the backend not filtering correctly?

This is an issue for both locations and datasets.  About the datasets filter David said:
> the problem with the projects download is that the measurements endpoint does not have a projects filter -- at some point we had talked about that we would just pass in all the locations (&location=123&location=124...) but I'll just add the filter since that seems to already be in place

PR #637, opened 2021-01-20
Improvements to download modal

danielfdsilva:


rwegener2 (changes requested):
I played with this a bit.  The LCS filter looks nice.  The info button is nice detail.  When I download data it looks like the filter isn't applying, however.  When I go to the modal, enter "United States" and try it twice - once with LCS selected and once with reference selected - the CSVs give me the same results.  It's possible that the API isn't returning correctly, but just post here if that is the case.

The pre-populating of locations that don't have a city works well (two I tried listed below). What seems to be not working not is pre-populating of locations that do have a country listed.  I put some examples below to check into.
### Not working Examples (includes country)
http://localhost:3000/#/location/62446
http://localhost:3000/#/location/62392
### Working Examples (no country)
http://localhost:3000/#/location/67437
http://localhost:3000/#/location/67127

## Summary
* card "[Daniel] add a filter for lcs vs reference data (`sensorType`) for the locations page Data Download modal": GUI is nice, check functionality for filtering
* #625: Complete in the original scope of the ticket.  New filter error with LCS (above) to be investigated
* #626: Original error solved; investigate new error

danielfdsilva:
@rwegener2 Pushed fixes for the pre-selecting bug and rebased from current `develop`

As for the `sensorType`, I see that the URL is being constructed correctly: `https://0jac6b9iac.execute-api.us-east-1.amazonaws.com/v2/measurements?format=csv&limit=66536&country=AD&sensorType[0]=low-cost%20sensor`. However, checking the docs I see that the `sensorType` filter is only available for `/locations`. This will need to be included in the `/measurements` as well. cc @bitner

rwegener2:
Thanks so much @danielfdsilva.  The fixes work really well to me.

I've pinged David about the API request.  So you know we have switched our API addresses to be located on OpenAQ's AWS account.  The new staging url is `https://b4qc1cqcm2.execute-api.us-east-1.amazonaws.com/`.  This URL does seem to have that sensorType filter but I think there is still a bug in it so I've still reached out to David to get that resolved.

danielfdsilva:
@rwegener2 the documentation for `/measurements` lists that `sensorType` as string, but for `/location` is an array.
Given that there are only 2 options, an array is not really useful, but I'd say that the endpoints should have the same api.

I can update this and only include this filter when only 1 option is selected, but ensuring a consistent api could make more sense. Let me know.

rwegener2:
Let's go with the string, as it is on measurements right now.  Standardizing between locations and measurements is a good idea; I'll add that to the list of backend requests for Bitner.

I was hoping to show this to the partner at the top of the hour (11am EST) so if there is time to merge by then that would be great.

rwegener2 (approved):
🎉 looks and works great

PR #638, opened 2021-01-20
Add check for locationIds

jvntf:
Mobile location pages were crashing. This is a quick fix, not sure if this param should actually be passed somewhere? @AliceR


AliceR (approved):
Thanks!

PR #646, opened 2021-01-21
Update/643 area

### LINKED ISSUE #643: Convert "Area" text to "City/Region"
We would like to switch the descriptor "Area" or "City" to "City/Region".  I have found places this needs to change:
1. in the download modal
2. In the Order By selection

If anyone sees other instances of City or Area then those can also be changed.  If not, this tasks is complete with those two.

jvntf:
closes #643 
Also changes area in compare location selector
Also fixes sort options on projects hub

### DIFF
--- a/app/assets/scripts/views/locations-hub/locations-hub.js
+++ b/app/assets/scripts/views/locations-hub/locations-hub.js
@@ -60,9 +60,7 @@ export default function LocationsHub({
     setFilters({
       // In the front end we are using param 'area', but this is
       // mapped to 'city' before getting sent to backend.
-      order_by:
-        query.order_by &&
-        query.order_by.split(',').map(opt => (opt === 'area' ? 'city' : opt)),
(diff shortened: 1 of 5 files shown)


rwegener2 (approved):
Nice, thanks for finding those additional instances of the old verbage

app/assets/scripts/views/locations-hub/locations-hub.js
```diff
@@ -58,7 +58,9 @@ export default function LocationsHub({
     setPage(() => getPage(query));
 
     setFilters({
-      order_by: query.order_by && query.order_by.split(','),
+      // In the front end we are using param 'area', but this is
+      // mapped to 'city' before getting sent to backend.
```
AliceR, line 62: Maybe we can consider the renaming on the backend as well, for better consistency? Or is there a specific reason it needs to be called city in the backend?

PR #651, opened 2021-01-22
Feature/648 analysis

### LINKED ISSUE #648: [Epic] Analysis Datasets
### Two types of data
EDF data comes in two types - raw data and analysis data.  Raw data is streaming data, it comes directly from the source.  Analysis data has been collected, grouped, and had some pretty specific processing done to it.  Functionally the difference lies in the fact that raw data can be uniquely identified with a sensor id and a timestamp.  Analysis data is averaged across what is currently a several days long window.

As a note all of the analysis data is also mobile data.

### How we are handling it
Because of the different nature of this data it has been decided that we will store that data separately in the database and label it with the flag `isAnalysis` = True.  It will still be available through the `measurements` endpoint and can be associated with locations or datasets, but the data will not be aggregated at the `averages` endpoint.

## Goal (Frontend)
To accommodate this data on the frontend we will make a few changes:
* label locations as `Analysis` or `Raw` in locations hub and dashboards
* label datasets as `Analysis` or `Raw` in datasets hub and dashboards
* create a new type of dataset dashboard for the Analysis-only datasets

## Details
**Update:** We are going to label the analysis data as such but leave the rest of the data with no alternative tag.

### label locations as `Analysis` in locations hub and dashboards
Adding a tag which says "Analysis" to the location card (orange) and also in the header of the dashboard (white).  These can go next to the other tags (stationary, community, reference, etc.)

**API**
`results` > `isAnalysis` returns a boolean
Analysis: https://b4qc1cqcm2.execute-api.us-east-1.amazonaws.com/v2/locations/61495 
Not analysis: https://b4qc1cqcm2.execute-api.us-east-1.amazonaws.com/v2/locations/62216

### label datasets as `Analysis` in datasets hub and dashboards
same as above

API
_not yet implemented_

### create a new type of dataset dashboard for the Analysis-only datasets
The dashboard for analysis datasets will not contain a time series, temporal coverage, or parameters components.  They will contain only the top three grey components and the mobile data map.

API
Can start working removing the components now?  Unfortunately there isn't a flag on projects right now to distinguish between the type of dataset.

jvntf:
I think we waiting on API before this can be fully tested but this should be good to go.

closes #648 648

### DIFF
--- a/app/assets/scripts/components/filter.js
+++ b/app/assets/scripts/components/filter.js
@@ -21,6 +21,8 @@ const defaultSelected = {
   source_type: [],
 };
 
+const procLevelOptions = ['Analysis'];
+
 const initFromLocation = ({
   countries,
   parameters,
@@ -30,13 +32,14 @@ const initFromLocation = ({
   manufacturer,
   mobility,
   entity,
+  procLevel,
 }) => {
   return {
     parameters: parameters ? parameters.split(',').map(Number) : [],
     countries: countries ? countries.split(',') : [],
     sources: sources ? sources.split(',') : [],
     order_by: order_by ? order_by.split(',') : [],
-
+    procLevel: procLevel,
     grade: grade,
     manufacturer: manufacturer,
     mobility: mobility,
@@ -46,7 +49,6 @@ const initFromLocation = ({
 export default function Filter({
   slug,
   by,
-
   fetchBaseData,
   orderByOptions,
   countries,
@@ -105,6 +107,15 @@ export default function Filter({
           manufacturer: query.manufacturer,
           mobility: query.mobility,
           entity: query.entity,
+        }));
+        break;
+      }
+
+      case 'procLevel': {
+        query.procLevel = value;
+        setSelected(prev => ({
+          ...prev,
+          procLevel: value,
         }));
         break;
       }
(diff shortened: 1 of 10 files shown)


rwegener2:
🎉 Thanks @jvntf!  David just finished the API update: https://b4qc1cqcm2.execute-api.us-east-1.amazonaws.com/v2/projects?isAnalysis=true

AliceR (approved):
Is it intentional that only raw datasets show up initially? I'd expect to see both raw and analysis datasets together, unless a filter is selected?

Also noting that the test is failing right now because /v2/countries?limit=1000 returns 422. 


Otherwise the code looks good to me.

rwegener2 (changes requested):
Thanks for pulling all of these together, @jvntf!

Two things I would change from my original comment now that I'm seeing it:
* Let's not label cards as "Raw".  If `isAnalysis` is false let's just not show any additional tag.  This can apply to both Locations and Datasets cards
* Let's remove the filter on the Datasets page.  I do like the idea but seeing as there are so few datasets right now it is pretty manageable to skim.

rwegener2:
To address @AliceR's comment I think it makes the most sense to always have the two types showing up together.

jvntf:
comments addressed, thanks for reviews @AliceR @rwegener2

rwegener2 (approved):

PR #659, opened 2021-01-25
Connect mobile data to time selection

AliceR:
This displays the mobile data points within the selected time range as blue, and those not within in grey.

Note: the property 'lastUpdated' is missing on the tiles for the mobile points, hence currently all points are grey, always.

### DIFF
--- a/app/assets/scripts/views/project/dashboard.js
+++ b/app/assets/scripts/views/project/dashboard.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import moment from 'moment';
 
 import DetailsCard from '../../components/dashboard/details-card';
 import LatestMeasurementsCard from '../../components/dashboard/lastest-measurements-card';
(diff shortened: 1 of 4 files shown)


app/assets/scripts/views/project/dashboard.js
```diff
@@ -46,8 +47,28 @@ function Dashboard({
           locationId={locationIds.length === 1 ? locationIds[0] : null}
           locationIds={locationIds}
           bbox={bbox}
-          firstUpdated={projectDates.start}
-          lastUpdated={projectDates.end}
+          dateRange={
+            dateRange
```
AliceR, line 51: the complexity here made me wonder if we should store the dateRange in state as {start, end} directly instead? I did not want to do all the related refactoring at this point, but it would maybe be a good idea?

jvntf (changes requested):
At time of test mobile location pages are crashing, see http://localhost:3000/#/location/218338

image

AliceR:
Thank you @jvntf , I forgot to add the dateRange prop to the card on the location dashboard. Should be fixed with 68fc70d

jvntf (changes requested):
Thanks for the fix @AliceR
In testing [here](http://localhost:3000/#/project/28957?dateRange=2014/12)
I see only gray dots when i select both full project and single month. Not sure if they color correctly on single day (not sure which days have data so couldn't test a specific day). Likely unrelated, but for the above location I see data points on the map and data in temporal coverage, but time series shows no data.

AliceR:
That's related to what I noted in the PR description: The property 'lastUpdated' is missing on the tiles for the mobile points, hence currently all points are grey, always.

At least points show up on the project, I get nothing on mobile location pages :/

jvntf:
Ah sorry i missed your note! 🙈 ok looks good then!

PR #662, opened 2021-01-25
Update url used for feedback

AliceR:
adding it to the config for convenient future updating.

- [x] TODO: clarify what should happen with the "Add new source form". I assume to stay as is?


jvntf (approved):

rwegener2:
@AliceR I couldn't find anywhere where "Suggest a Source" redirected to the LCS Feedback form.  Do you have a specific page you were looking at? For me Suggest a Source always redirected here https://docs.google.com/forms/d/e/1FAIpQLSdcmn3qxoM69XIjJ3Pl4cyO6Ffk8TT40NC_XizS6snAQ7Z8_A/viewform

I do think that the path we want to take is for Suggest a Source to always link to the url above ^ and the Share Feedback button to always link to the new URL https://forms.gle/rqsAPDSebNTG8PsJ8

AliceR:
Ok, sounds good. 

I extracted the other url into the config file as well: https://github.com/openaq/openaq.org/pull/662/files#diff-5cf01a8d749dcfcd413d9ca26ceaa7f8fa72ff8f3d9e60ceb5afe7b0d65f9deb

PR #665, opened 2021-01-25
Fix styling

jvntf:
Fixes https://github.com/openaq/openaq.org/projects/1#card-53118492
Thanks @danielfdsilva for the fix


app/assets/scripts/components/sources-list.js
```diff
@@ -44,7 +44,7 @@ export default function SourcesList({ sources, showReadme, showOrganisation }) {
                   rel="noreferrer noopener"
                   target="_blank"
                 >
-                  Technical readme
+                  Technical readm
```
AliceR, line 47: that looks accidental

AliceR (changes requested):
Definitely better, but still weird:
Screenshot 2021-01-25 at 20 49 59

jvntf:
@AliceR added a break-word style for the header. not ideal but not sure what else to do about that.

AliceR (approved):

PR #666, opened 2021-01-25
Update/country pagination

necoline:
- adds pagination to country page


jvntf (changes requested):
I think this needs some padding, other than that looks good. (Circles are on line, on other pages there is some space)
image

AliceR (approved):
@necoline this is great! 

Hope you don't mind that I moved around some code, to keeps the page rendered while the paginated list is updated and also avoiding to refetch the country data on every page selection. Thanks for making such adjustments easy with encapsulated logic! 🙌 

Oh and I added sort and order_by parameters to the api request, to make the locations appear in a bit more systematic order.

PR #668, opened 2021-01-25
Set max zoom on fit bounds

AliceR:
avoid zooming in too far when there is only a tiny bbox (1 single measurement, or not at all moving).

E.g. http://localhost:3000/#/location/45152 - were the _accurate_ bbox provided by the locations object is way smaller than the generalized 30m bounds provided with the tiles


jvntf (approved):
Not sure I understand your comment about having super tiny bboxes and what to do in that case. But otherwise, this works as described.

PR #671, opened 2021-01-25
Update/669 rounding

### LINKED ISSUE #669: Unit rounding
## Goal
The Latest Measurements component at the top can look a little sloppy (and convey undue certainty) when very large numbers are rounded to several decimal places.  

Instead lets show:
* any number with a ~value > 2~ value < 2: round to two decimal places
* any number with a value >= 2: round to the nearest whole number

## Details
These changes will be relevant on both the projects dashboards (Ex. Houston Mobile https://openaq-staging.org/#/project/28974) and the locations dashboards (Ex. https://openaq-staging.org/#/location/61507)

## Additional
This is most important in the Latest Measurements component, but if possible this should apply to all numbers on the site.  This includes:
* time series hover text on dashboards
* parameters table values
* latest reading in the map popups

jvntf:
closes #669 
- [x] latest measurements
- [x] time series hover text on dashboards

- [x] parameters table values

- [x] latest reading in the map popups

### DIFF
--- a/app/assets/scripts/components/dashboard/lastest-measurements-card.js
+++ b/app/assets/scripts/components/dashboard/lastest-measurements-card.js
@@ -31,7 +31,7 @@ export default function LatestMeasurementsCard({ parameters }) {
                   {o.displayName}
                 </CardSubtitle>
                 <HighlightText className="card__highlight-text" size="medium">
-                  {round(o.lastValue, 2)}
+                  {round(o.lastValue, o.lastValue < 2 ? 2 : 0)}
                 </HighlightText>
 
                 <strong>{o.unit}</strong>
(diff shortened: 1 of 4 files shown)


rwegener2:
The two things that are done right now look super nice.  I checked out several pages, locations and datasets, and the numbers in the Latest Measurements section look good.

rwegener2 (approved):
Number rounding looks great to me.

I did notice this on the world map.  The expected popup is the location I clicked on and the json text is the point that my cursor is hovering over.  Is this something caused in this PR or introduced somewhere else?

Screen Shot 2021-01-25 at 9 22 37 PM

AliceR:
@rwegener2 the hover is something that only shows up when running locally. I use it to debug properties provided with the map tiles.

AliceR (approved):

Issue #667, opened 2021-01-25
Incorporate analysis visual into maps


rwegener2:
## Goal 
Distinguish the Analysis data on the map somehow

### Open Questions
* How prominently should we distinguish (a lot? -> triangles; quite a bit? -> black icons; not a lot? -> just in popup)
If needed let's remember to update the map legend

rwegener2:
The decision moving forward is not to highly emphasize the analysis data on the map.  So for our solution this task will just involve adding the orange "Analysis" tag to the popup.

PR #672, opened 2021-01-25
add card tag

necoline:
#667
To test check out nearby locations at `location/61484` with the BC parameter


AliceR (approved):
Despite your instructions I did not find any example in the UI, but the code looks good!

PR #673, opened 2021-01-25
Use symbols for node layer, highlight selected

AliceR:
Reusing symbols from measurement layer. Adding a outline color when selected.

### DIFF
--- a/app/assets/scripts/components/map/node-layer.js
+++ b/app/assets/scripts/components/map/node-layer.js
@@ -4,7 +4,11 @@ import PropTypes from 'prop-types';
 import { useRouteMatch } from 'react-router-dom';
 import mapbox from 'mapbox-gl';
 
-import { defaultColor, unusedBorderColor } from '../../utils/colors';
+import {
+  defaultColor,
+  highlightColor,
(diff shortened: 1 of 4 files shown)


app/assets/scripts/components/map/node-layer.js
```diff
@@ -27,106 +25,89 @@ export default function NodeLayer({
...
 
+  useEffect(() => {
+    map.setPaintProperty(`${activeParameter}-node-outline`, 'icon-color', [
+      'case',
+      [
+        'in',
+        ['number', ['get', 'locationId']],
+        ['literal', allSelectedLocations],
+      ],
+      unusedBorderColor,
```
necoline, line 106: I would recommend a different color, I think this is a little hard to recognize against the blue

necoline (approved):

rwegener2 (approved):
The oakland locations show up great (project/29097) 🎉 

I can't say I can really see the change when selected, though.  Maybe if the symbol got bigger?  Or just a bigger outline?

AliceR:
You both are absolutely right, I am going to invest a few more minutes into color selection 🎨

PR #680, opened 2021-01-27
Update country map + other map additions

### LINKED ISSUE #677: generalize countries map
## Goal
After discussing with the group the goal of the maps on the countries pages is not to show most recent values but to highlight where data is available.  Because of this it was decided it would be more important to show all the locations, regardless of parameter, than just show stations with PM2.5.

## Plan
What is the feasibility of showing all the dots on the map in black or something and showing all locations not matter what parameters they report?  That would be a good option but folks have also agreed that this could be a clarifying hover text instead.

### Commitment
This would make some people happy but isn't the highest priority item.  It should be time scoped to 1 hour, maybe an hour and half max.

AliceR:
Close #677 by adding a locations layer without measurement values.

While working on maps, I also added some more fixes:
- Update legend to render without parameters
- Update popup to render select button independent from api
- Fix location highlight on nearby locations

### DIFF
--- a/app/assets/scripts/components/map/location-layer.js
+++ b/app/assets/scripts/components/map/location-layer.js
@@ -3,11 +3,13 @@ import PropTypes from 'prop-types';
 
 import { ParameterContext } from '../../context/parameter-context';
 import {
-  coloredCircleRadius,
-  selectCircleRadius,
-  selectShadowCircleRadius,
+  iconMatch,
+  coloredSymbolSize,
+  selectSymbolSize,
+  selectShadowSymbolSize,
 } from '../../utils/map-settings';
 import { getFillExpression, defaultColor } from '../../utils/colors';
+import { square, circle } from './symbols';
 
 export default function LocationLayer({
   activeParameter,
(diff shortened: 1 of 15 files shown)


app/assets/scripts/components/map/location-layer.js
```diff
@@ -51,13 +57,12 @@ export default function LocationLayer({
...
-        filter: ['in', 'locationId', ['literal', locationIds]],
+        filter: ['in', ['get', 'locationId'], ['literal', locationIds]],
         type: 'circle',
         paint: {
-          'circle-color': {
-            property: 'lastValue',
-            stops: generateColorStops(activeParameter),
-          },
+          'circle-color': isCore(activeParameter)
+            ? getFillExpression(activeParameter)
```
necoline, line 64: I'm seeing a strange behavior here. When this is a core parameter and a reference grade sensor, I see both square and circle. The example I'm looking at is `http://localhost:3000/#/location/62302` nearby locations with an active parameter of `PM2.5`

necoline (changes requested):

app/assets/scripts/components/map/location-layer.js
AliceR: good catch! I did not update the highlight location layer to use the new symbology

rwegener2:
The new layer looks nice and will be much appreciated.

I did have an issue with my popups not rendering:
Screen Shot 2021-01-27 at 8 19 17 AM

rwegener2:
@AliceR  pointed out that the issue in my previous comment was a popup that shows only the debug popup (which only is available in development mode).  Clicking on the points the popup works great

rwegener2 (approved):

necoline (approved):

PR #687, opened 2021-01-27
If page not found, redirect to home

AliceR:
I realize this project has no 404 page? Ideally, a 404 screen would be designed and implemented, but for now we can simply redirect to '/' (the home view), when the url doesn't match an expected path.

This happened with http://localhost:3000/#/locations/60544 (where a valid url would be http://localhost:3000/#/location/60544, note the s in locations)

Additionally, the location page would crash if the api returns no data for the location id. Now, we display an error message instead.

### DIFF
--- a/app/assets/scripts/views/app.js
+++ b/app/assets/scripts/views/app.js
@@ -2,7 +2,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { withRouter, matchPath } from 'react-router-dom';
+import { withRouter, matchPath, Redirect } from 'react-router-dom';
 import c from 'classnames';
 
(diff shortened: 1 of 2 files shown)


necoline (approved):

AliceR:
accessing /location still causes a total breakdown, need to fix this too

rwegener2:
The links you gave @AliceR did redirect me to the home page.  I also tested the ID 36083 (which doesn't exist https://b4qc1cqcm2.execute-api.us-east-1.amazonaws.com/v2/locations/36083) and it gave me the "Uh oh, something went wrong." page, which I think is also a totally good option. Much better than crashing.

rwegener2 (approved):

Issue #678, opened 2021-01-27
Time series daily project lifetime


rwegener2:
## Goal
Provide a higher data resolution to the users on the default view of the dashboard page.

## Details 
Time series component:
We would like to show the hourly average of all the data across the entire lifetime of the dataset.  Right now I think by default it might show the latest week

necoline:
I thought this was our plan so as to reduce expensive calculations. Maybe I'm remembering this wrong
cc: @bitner

rwegener2:
@necoline David and I chatted about this yesterday.  We are keeping the binning at daily, which means that even two years of daily data is 730 points.

I am a little concerned about crashing if a sensor has years and years of data, so we could add a two year max to this.

PR #688, opened 2021-01-27
adds default date range - last 2 years

necoline:
addresses #678
good example to test `/location/21`
- adds a 2 year date range default on time series average request.
- increases limit to 1000

**Unresolved: We are not getting averages for the whole collection period from the backend. 
- project ex: 2 year lifespan only returns 1 year of data 

https://u50g7n0cbj.execute-api.us-east-1.amazonaws.com/v2/averages?temporal=moy&parameter=co2&spatial=project&project=richmond_beaco2n

- location ex: 6 year lifespan only returns 2 years of data
https://u50g7n0cbj.execute-api.us-east-1.amazonaws.com/v2/averages?parameter=pm25&temporal=day&location=21&spatial=location

### DIFF
--- a/app/assets/scripts/components/dashboard/time-series-card.js
+++ b/app/assets/scripts/components/dashboard/time-series-card.js
@@ -36,6 +36,7 @@ const defaultState = {
 export default function TimeSeriesCard({
   locationId,
   projectId,
+  lastUpdated,
   parameters,
   prefetchedData,
   dateRange,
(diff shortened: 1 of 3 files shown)


AliceR (commented):
This seems to work, but I can't make sense of all timelines. For example http://localhost:3000/#/location/21 -> why does it only go back until 2020/10/14 if the collection dates range back to 2015? I'd expect there to be 3 years based on your code, or 2 years based on api limitation?
Also I find our dateRange setup somewhat confusing, but this is for a later time...

app/assets/scripts/components/dashboard/time-series-card.js
```diff
@@ -76,6 +77,10 @@ export default function TimeSeriesCard({
   const fetchData = () => {
     setState(state => ({ ...state, fetching: true, error: null }));
 
+    // get date 2 years prior to last updated
+    let defaultStartDate = new Date();
+    defaultStartDate.setFullYear(new Date(lastUpdated).getFullYear() - 3);
```
AliceR, line 82: should this be `- 2`?

AliceR (approved):

PR #690, opened 2021-01-27
Fix page

### LINKED ISSUE #679: search - reset to page 1 when filters change
### From partner feedback
the page number doesn’t reset if you clicked to a later page and then do a new search from that page.

For example, this url works:  https://openaq-staging.org/#/locations?mobility=Mobile

Now scroll to bottom and click the page 2 gets you to this page:  https://openaq-staging.org/#/locations?mobility=Mobile&page=2

From there, select NO2 as the parameter as additional filter, takes you to this page:  https://openaq-staging.org/#/locations?mobility=Mobile&page=2&parameters=7

No locations show up on that page because there are only 6 results with NO2 so page 2 shows up as empty.

**Can you get the page to go back to the beginning when the filters change?**

Page shown when the filters are reset:  https://openaq-staging.org/#/locations?mobility=Mobile&parameters=7

jvntf:
closes #679

### DIFF
--- a/app/assets/scripts/components/card/card.js
+++ b/app/assets/scripts/components/card/card.js
@@ -65,7 +65,8 @@ export default function Card({
   noBodyStyle,
   titleInfo,
 }) {
-  console.log(tags);
+
+    console.log(tags);
   return (
     <article data-cy={`${id}-card`} className={c('card', className)}>
       <CardContents
--- a/app/assets/scripts/components/card/card.js
+++ b/app/assets/scripts/components/card/card.js
@@ -65,8 +65,6 @@ export default function Card({
   noBodyStyle,
   titleInfo,
 }) {
-
-    console.log(tags);
   return (
     <article data-cy={`${id}-card`} className={c('card', className)}>
       <CardContents
--- a/app/assets/scripts/components/card/card.js
+++ b/app/assets/scripts/components/card/card.js
@@ -65,6 +65,7 @@ export default function Card({
   noBodyStyle,
   titleInfo,
 }) {
+  console.log(tags);
   return (
     <article data-cy={`${id}-card`} className={c('card', className)}>
       <CardContents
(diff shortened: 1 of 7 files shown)


app/assets/scripts/components/card/card.js
```diff
@@ -65,6 +65,7 @@ export default function Card({
   noBodyStyle,
   titleInfo,
 }) {
+  console.log(tags);
```
AliceR, line 68:
```suggestion

    console.log(tags);
```

app/assets/scripts/views/locations-hub/locations-hub.js
```diff
@@ -55,7 +55,15 @@ export default function LocationsHub({
     let query = qs.parse(location.search, {
       ignoreQueryPrefix: true,
     });
-    setPage(() => getPage(query));
+    setPage(prev => {
+      const cur = getPage(query);
+      if (prev === cur) {
+        //This means that only filtes have updated
```
AliceR, line 61: 👍 thanks for the explanation! Valuable comment

cypress/integration/locations-hub.spec.js
```diff
@@ -90,4 +90,22 @@ describe('The Locations Hub', () => {
     cy.get('[data-cy=location-card-detail-label]').contains('Parameters');
     cy.get('[data-cy=location-card-detail-label]').contains('Source');
   });
+
+  it('Pagination is reset on filter change', () => {
```
AliceR, line 94: 🎉 ❤️ 🥳

rwegener2 (approved):
🎉

AliceR (changes requested):
I ran into a loop when trying to filter for ow cost sensors. I'll get you more info soon

jvntf:
@AliceR loop fixed

necoline (approved):

AliceR:
But the page click does not seem to work? It always resets me back to page 1 when I click the pagination button

AliceR:
@jvntf when there is a space in the filter parameter name (`Low-Cost Sensor`), the browser updates to url to `Low-Cost%20Sensor` which triggers the effect while prev and cur page are the same, hence it goes to page 1 again 🙈

AliceR (approved):

necoline (approved):

PR #694, opened 2021-01-27
Fix country page

AliceR:


rwegener2 (approved):
No crashing on my end.  Map is showing up as the not-parameter specific version
http://localhost:3000/#/countries/US
http://localhost:3000/#/countries/CN

🚀

necoline (approved):

Issue #692, opened 2021-01-27
More text updates


rwegener2:
**Dashboard: Temporal Coverage**
- [ ] show the x axis values in increasing order (hour of day shows 1:00 on the left side, then 3:00 and so on.  Day of week starts with Monday on left, then Tuesday and so on.  Year starts with January on the left.

**Mobile Dashboard pages**
- [ ] "Values for all __ stations": remove the word "all" and change the word "stations" to "nodes"

**Locations Dashboards**
- [ ] "Nearby Locations in __, __" I think should just say "Nearby Locations".  We have such intermittent city/country info that it doesn't add much to have it there.

PR #696, opened 2021-01-27
Update wording

AliceR:
Contribute to #692

### DIFF
--- a/app/assets/scripts/components/dashboard/temporal-coverage-card.js
+++ b/app/assets/scripts/components/dashboard/temporal-coverage-card.js
@@ -120,6 +120,8 @@ export default function TemporalCoverageCard({
       );
 
       let query = {
+        sort: 'asc',
+        order_by: temporal,
         temporal,
         parameter: activeTab.id,
         spatial,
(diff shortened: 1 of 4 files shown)


rwegener2:
Not sure which of the tasks in the ticket were intentionally targeted.  
#### I do see nice looking updates to:
* Nearby Locations

#### I don't see:
* temporal coverage x axis sorting
http://localhost:3000/#/location/66959

#### Comment
* mobile dashboard - station -> node 
I think a 🐍 got in here because we are showing information for nodess (extra "s").  I'm guessing the base word should be "node" instead of "nodes" that the "s" gets added to.
"2 nodess": http://localhost:3000/#/project/28974
"1 nodes": http://localhost:3000/#/project/28958

AliceR:
Thanks @rwegener2 , the snake is fixed.

I am sorting out the sorting with @bitner, the api doesn't seem to sort the way it should. but the query parameters as precondition on the frontend are placed with the latest commit  43c6c44

rwegener2 (approved):

PR #721, opened 2021-02-15
Parameters table

### LINKED ISSUE #717: parameters table not updating with time change
## Overview
It looks like the parameters table on the dashboards pages don't appear to be changing values when a different date is selected.  These values should be changing.

## Details
The values in the parameters table are all the same for the following links:
https://openaq.org/#/location/62806?dateRange=2021/1/30 
https://openaq.org/#/location/62806?dateRange=2021/1
https://openaq.org/#/location/62806

### Pondering
It's possible this could be a rounding issue, where the values just aren't changing significantly enough to show on the interface.  I think it's worth investigating though.

AliceR:
Rename the measurands card to `parameters card` in order to reflect the new naming convention. 

If there is a date range selected, fetch the data for the parameters table from /averages instead of using the summary provided by /locations:id.

Close #717

### DIFF
--- a/app/assets/scripts/components/dashboard/parameters-card.js
+++ b/app/assets/scripts/components/dashboard/parameters-card.js
@@ -37,6 +37,7 @@ const tableHeaders = [
 export default function ParametersCard({
   parameters,
   locationId,
+  projectId,
   dateRange,
   titleInfo,
 }) {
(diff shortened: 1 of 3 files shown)


jvntf (approved):
Looks good!

rwegener2 (approved):

PR #725, opened 2021-02-22
Display other related parameters as measurement layer

### LINKED ISSUE #720: world map - displaying parameters with different units
## Overview
In the old openaq.org website we think that there were more locations that were displayed on the world map because for some parameters the map was displaying parameters with different units together.  We would like to get more locations to show up on the map again, but the details of how need investigation.

## Details
In our new backend system pollutants with different units are considered to be completely unique.  For the purposes of the map, however, we still want both units of measurement to show up as though they were the same parameter.  It isn't clear if previously the non-standard units were all showing up grey, if they had a unique color scale, or something else.

### Parameters
In the table we have listed the "Displayed Parameter" which is the parameter that is currently being displayed in the world map.  The "Add Parameter" is the one that we would like to add to the world map together with the Displayed Parameter.
|  displayed parameter (ID) | add parameter (ID) | conversion |
|---|---|---|
| CO (8) | CO mass (4)  | 0.000904379 | 
|  NO2 (7) | NO2 mass (5)  | 0.000550492 | 
|  O3 (10) |  O3 mass (3) | 0.000395666 |
|  SO2 (9)| SO2 mass (6)  | 0.000527554 |

The name of the parameter on the legend does not have to change.

## Tasks
* Task 1 of this ticket is to do a preliminary look to see if we can tell what was previously happening with the map.  This should be time capped, though, to maybe 30-45 minutes so as to not go down a rabbit hole.
* Task 2 will be to implement adding locations the non-standard units as either greyed out points or set up a system giving them their own color scale.

Between tasks 1 and 2 we will need to consult OpenAQ to see how they want to proceed in task 2.

AliceR:
Close #720 

When selecting a parameter such as CO we can now see locations of both units (µg/m³ and ppm) displayed together on the world map. The underlying values are not converted, only the max. color value is that is used to set the symbol color.
The grouping of parameters is done using their `name` property.

The api will include the converted max color value to the /parameters response. Until this value is included for those 4 parameters we have the conversion rate (magic numbers from old code), the locations fall back to use OpenQA-blue.


rwegener2 (approved):
🎉

app/assets/scripts/utils/map-settings.js
```diff
@@ -106,47 +106,3 @@ export const locationShadowCircleRadius = {
...
-  let value = parameter.value;
-  let p = parameter.parameter;
-  if (
-    Object.keys(parameterConversion).indexOf(p) !== -1 &&
-    parameter.unit !== parameterUnit[p]
-  ) {
-    value *= parameterConversion[p];
-  }
-  return value;
-}
```
necoline, line 152: 🎉

necoline (approved):
looks good! I found a reference to one of the deleted home pages which caused issues loading the home page. Check out my commit to make sure it looks good before merging

AliceR:
Thanks for pointing out the issue, @necoline! I did not find the commit you mention, but added d52fe82 to fix the home page charts. Also double checked for other references to the deleted code, should be good now.

PR #729, opened 2021-02-24
Parcel

necoline:


AliceR:
Some notes on the process of migrating to Parcel:
- we had to update the way images are handled. They used to be copied over to /dist in a gulp task. Now they need to be imported. We might want to look into image optimization.
- logo and favicon are coming over from openaq-design-system. We can either copy them into this repo to import, or import them from node_modules
- the previous app/assets/content/content.json was generated based on .md files in a gulp task. We looked into markdown to json parser but could not come up with a satisfying solution. For now, the .md files are replaced with the parsed json. We should look into this, mainly to make content maintenance as easy as possible.

- [ ] TODO: href to "/assets/files..." needs to be fixed

AliceR:
Screenshot 2021-02-25 at 17 28 04
The current bundle, visualised as a treemap. It indicates areas we need to improve:

- files like pdf's, images
- use only one date library instead of both moment.js and date-fns
- mapbox gl is huge but we can't change that
- ... ?

AliceR:
from the test results it looks like it is not yet finding the API env var - I did ask Sruti to add them, it _should_ be there

necoline:
Awesome, was just about to point that out

AliceR:
I forgot that I need to reference the secrets as env vars in the github actions, done in 2d3713c

PR #730, opened 2021-02-24
Add lighthouse workflow

jvntf:
Add generic lighthouse workflow


necoline:
maybe it's time to upgrade node to 12?
https://github.com/GoogleChrome/lighthouse/issues/11854

AliceR:
I wonder if it would be easier if applied onto #729 ? At least no hassle with gulp

necoline:
great job Jeevan! It's great to see these tests results. These reports aren't that surprising. I think the next step will be documenting how to develop with lighthouse reports. I don't the goal is make these general health checks rather than development blockers

jvntf:
yes i agree, perhaps this app should be added to the repo to show reports? https://github.com/apps/lighthouse-ci

Lighthouse action should ideally not show as a failed test.

necoline:
Let's add it in! I think we should be able to customize how it runs in the config file. I took an initial stab at removing our failing tests. The idea conveyed in the docs is that these assertions can acts as a todo list for improving the site, once addressed the failures highlight regressions in the code

jvntf:
@necoline @AliceR do you know if we can safely remove the `npm install lhci` and `npm rebuild node-sass` from the workflow? im not sure if the former needs to be installed every time. the latter was needed after changing the node version, hopefully not that its been run it doesnt need to be included?

necoline:
I'm not sure about installing `lhci` but it seems reasonable to remove the `node-sass rebuild`

.env.development
```diff
@@ -0,0 +1,15 @@
+# Environment variables used by the app
```
AliceR, line 1: ups, these were not meant to be committed

necoline: 😬

PR #828, opened 2023-05-24
Added Ambassadors cards, photos and Ambassadors section to about/people

majesticio:
Has socials, country / city, bio. Made a new ambassador-card.html in themes/layouts/about/people


russbiggs:
looks good to me, just needs an update to the decap cms config.yml to include the new 'group' option

https://github.com/openaq/openaq.org/blob/main/static/admin/config.yml#L156

majesticio:
> looks good to me, just needs an update to the decap cms config.yml to include the new 'group' option
> 
> https://github.com/openaq/openaq.org/blob/main/static/admin/config.yml#L156

Oh yeah, thanks for pointing that out. Should be ok now.

russbiggs (approved):

